PR #4, opened 2022-08-01
IE-0004: Access to data files embedded in blorbs

curiousdannii:
- Proposal: [IE-0004](https://github.com/ganelson/inform-evolution/blob/main/proposals/0004-using-data-files-in-blorbs.md)
- Authors: Andrew Plotkin and Graham Nelson
- Language feature name: None
- Status: Draft
- Related proposals: None
- Implementation: None

## Summary

The ability to read binary data files included in a blorbed release of a game.

### DIFF
--- a/proposals/0004-using-data-files-in-blorbs.md
+++ b/proposals/0004-using-data-files-in-blorbs.md
@@ -1,6 +1,7 @@
 # (IE-0004) Access to data files embedded in blorbs
 
 * Proposal: [IE-0004](0004-using-data-files-in-blorbs.md)
+* Reference link: [#4](https://github.com/ganelson/inform-evolution/pull/4)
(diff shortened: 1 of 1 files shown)


erkyrath:
> We will also allow `internal text` and `internal form`

I'll have to re-read the blorb docs, but I think `internal form` is not needed. It's the automatic result of using `internal data` when the data file is IFF-formatted (e.g. an AIFF or Quetzal file).

zedlopez:
I would assume this is implausible and that there is no provision for non-objects to have a kind hierarchy. But it would be kind of nice if one could have file variables that could be asigned *either of* an external file or an internal file and one could write generic phrases that could handle either, facilitating things like: "use a user-provided file if there is one; otherwise fall back to included file".

Will internal file and internal text be able to have properties (like Figure and Sound) or not (like external files)?

PR #6, opened 2022-08-01
IE-0006: New annotations for I6 syntax

curiousdannii:
- Proposal: [IE-0006](https://github.com/ganelson/inform-evolution/blob/main/proposals/0006-i6-syntax-annotations.md)
- Authors: Graham Nelson, David Kinder, and Andrew Plotkin
- Language feature name: None
- Status: Draft
- Related proposals: https://github.com/DavidKinder/Inform6/issues/189.
- Implementation: None

## Summary

A standard way to mark up Inform 6 syntax to provide annotations to help tools with linking or compilation, but which do not change its basic meaning.

### DIFF
--- a/proposals/0006-i6-syntax-annotations.md
+++ b/proposals/0006-i6-syntax-annotations.md
@@ -1,6 +1,7 @@
 # (IE-0006) New annotations for I6 syntax
 
 * Proposal: [IE-0006](0006-i6-syntax-annotations.md)
+* Reference link: [#6](https://github.com/ganelson/inform-evolution/pull/6)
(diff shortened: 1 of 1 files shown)


curiousdannii:
If you replace a function then there should be a way to refer to the original version, so that you can create wrappers without needing to duplicate the original code.

zedlopez:
I'm going to call attention to [I7-2269](https://inform7.atlassian.net/browse/I7-2269) regarding I6 inclusions that include floating point literals having their +'s stripped when going from Inter->I6 and becoming uncompilable, just in case something about this new syntax would complicate fixing that.

ganelson:
Not any more: I7-2269 is fixed.

PR #8, opened 2022-08-01
IE-0008: Faster memory allocation

curiousdannii:
- Proposal: [IE-0008](https://github.com/ganelson/inform-evolution/blob/main/proposals/0008-faster-memory-allocation.md)
- Authors: Dannii Willis
- Language feature name: None
- Status: Draft
- Related proposals: None
- Implementation: None

## Summary

Faster runtime access to flexibly-sized data like texts and lists.

### DIFF
--- a/proposals/0008-faster-memory-allocation.md
+++ b/proposals/0008-faster-memory-allocation.md
@@ -1,6 +1,7 @@
 # (IE-0008) Faster memory allocation
 
 * Proposal: [IE-0008](0008-faster-memory-allocation.md)
+* Reference link: [#8](https://github.com/ganelson/inform-evolution/pull/8)
 * Authors: Dannii Willis
 * Language feature name: None
 * Status: Draft


erkyrath:
> Glulx is very reliably implemented and we can have confidence that its malloc opcode will be efficient.

Optimist `:)`

I am reasonably confident that the malloc opcode is implemented reliably. But I suspect there will be a round of patching it up to be efficient enough for widespread use.

curiousdannii:
Even if it's quite unperformant, it's going to be better than Inform implementing its own malloc system as it does now!

I had originally suggested inlining most of the calls to `BlkValueRead`/`BlkValueWrite`, but Graham would prefer to keep them, and hinted that Inter might be able to handle inlining these calls. That would be fantastic!

Maybe one way to try to help keep things working in Z-Code would be to be more particular about how big a block is allocated for lists/texts? Ie, never over-allocate to begin with, only increase to the next smallest size block, and immediately decrease in size if the required length ever drops. If I'm reading the code correctly, the current list template never shrinks lists? So a change like that might go a long way to preventing unnecessary fragmentation.

PR #9, opened 2022-08-01
IE-0009: Dialogue Sections

curiousdannii:
- Proposal: [IE-0009](https://github.com/ganelson/inform-evolution/blob/main/proposals/0009-dialogue-sections.md)
- Authors: Emily Short and Graham Nelson
- Language feature name: dialogue
- Status: Draft
- Related proposals: [IE-0010](https://github.com/ganelson/inform-evolution/blob/main/proposals/0010-concepts.md)
- Implementation: None

## Summary

An extensible system for dialogue, combining natural-language playtext in "dialogue sections" of the source text with an active runtime component, the "director", implemented with a new kit, `DialogueKit`.

### DIFF
--- a/proposals/0009-dialogue-sections.md
+++ b/proposals/0009-dialogue-sections.md
@@ -1,6 +1,7 @@
 # (IE-0009) Dialogue Sections
 
 * Proposal: [IE-0009](0009-dialogue-sections.md)
+* Reference link: [#9](https://github.com/ganelson/inform-evolution/pull/9)
 * Authors: Emily Short and Graham Nelson
 * Language feature name: `dialogue`
 * Status: Draft


zedlopez:
The auto-inclusion of DialogueKit when there's a dialogue section confounds inbuild's build-needs, build-locate, and graph features. I'm guessing it would be a big pain to try to remedy this; if that's true, I'd just suggest that inbuild's documentation call attention to this.

It seems like DialogueKit isn't slightly useful without WorldModelKit (and WorldModelKit isn't slightly useful without CommandParserKit), so would it make sense for DialogueKit's metadata to specify that it `needs` CommandParserKit ?

PR #13, opened 2022-10-11
IE-0013: Annotations for kit linking

curiousdannii:
* Proposal: [IE-0013](https://github.com/ganelson/inform-evolution/blob/main/proposals/0013-annotations-for-kit-linking.md)
* Authors: Graham Nelson
* Language feature name: --
* Status: Draft
* Related proposals: #6
* Implementation: Partly implemented but unreleased

## Summary

A simple set of annotations, in the sense of #0006, so that directives in the source code for kits or in `Include (- ... -)` code can be marked up in ways which affect how they are linked.

### DIFF
--- a/proposals/0013-annotations-for-kit-linking.md
+++ b/proposals/0013-annotations-for-kit-linking.md
@@ -1,7 +1,7 @@
 # (IE-0013) Annotations for kit linking
 
 * Proposal: [IE-0013](0013-annotations-for-kit-linking.md)
-* Discussion PR link: [#13](https://github.com/ganelson/inform-evolution/pull/13)
+* Discussion PR link: [#13](https://github.com/ganelson/inform-evolution/pull/13) 
 * Authors: Graham Nelson
 * Language feature name: --
 * Status: Draft


curiousdannii:
When replacing a function there should be a way to refer to the old version so that functions can be wrapped. Possibly that would make sense for constants too?

uecasm:
I don't like the idea of kits being able to hide identifiers.  Part of the point of extensions and customisation is that kits are often wrong about what things are "done" and don't need further customisation.  The power needs to be in the hands of the story author.  Namespaces are fine, though, just not making things private.

Isn't a better solution for the originally posed problem a method to define a dependency order between kits, such that you know that kit B is trying to replace kit A's definition because it explicitly says so?  Extensions do this via `Include` and via `replacing` (though the latter is quite clunky anyway); something similar could be used for the kits.

Having said that, I'm not entirely convinced the idea of using kits to get I6 code out of extensions is a good one -- unless there's a syntax for embedding a "kit" within an extension (not unlike the I6 code that's already there), you will end up with separate files that can become desynchronised and break unexpectedly because you have the wrong versions of things.

PR #15, opened 2023-01-17
IE-0015: World model enforcement

zedlopez:
- Proposal: [IE-0015](https://github.com/ganelson/inform-evolution/blob/main/proposals/0015-world-model-enforcement.md)
- Authors: Graham Nelson
- Language feature name: --
- Status: Draft
- Related proposals: --
- Implementation: Experiments in progress

## Summary

WorldModelKit and the Standard Rules, together with support hard-wired into the if module of the Inform compiler, collectively provide a "world model". At present it is not as consistent or as well-enforced as we would like, and this leads to unexpected and unhelpful behaviour. IE-0015 gathers up a set of changes to improve (though not fully fix) matters in one place, rather than leaving these in scattered responses to bug reports.

### DIFF
--- a/proposals/0015-world-model-enforcement.md
+++ b/proposals/0015-world-model-enforcement.md
@@ -1,7 +1,7 @@
 # (IE-0015) World model enforcement
 
 * Proposal: [IE-0015](0015-world-model-enforcement.md)
-* Discussion PR link: [#15](https://github.com/ganelson/inform-evolution/pull/15)  
+* Discussion PR link: [#15](https://github.com/ganelson/inform-evolution/pull/15)
 * Authors: Graham Nelson
 * Language feature name: --
 * Status: Draft


zedlopez:
As noted in [I7-2211](https://inform7.atlassian.net/browse/I7-2211) the world-model violations go beyond holding, wearing, and carrying; they also apply to supporting and containment. Is discussing those, also, in scope for this proposal?

The current proposal gives us different meanings of `hold` for each of assertions, now-phrases, and conditionals, all of which differ from the implied "hold" of both `holder of` and `first thing held by`.

- In assertions, things continue as today: only persons can hold and it means carrying.
- In now-phrases, only people, containers, and supporters can hold, and it means carrying, containment, or supporting, respectively.
- In conditionals, it means the same 5 relations `holder of` covers, except for always being false for non-things.

We're left without a verb corresponding to checking whether one region is the direct parent of another; that would have to be checked with `holder of`.

> in 10.1, neither of `X holds Y` and the `holder of Y is X` implied the other.

They come *really* close, though, the sole exceptions being when X is a non-person quantifier or a description of objects that's not a description of people.

I'm making a plea for having `hold` in conditionals continue to have its current meaning (but for changing the exception noted above), i.e., for it to correspond to `holder of`, bringing us from 5 different meanings to 4.

As I noted in [I7-2220](https://inform7.atlassian.net/browse/I7-2220), this parallels `enclose` (except for some edge-cases regarding floating objects and regions with rooms). In its introduction to enclosure, Writing with Inform invokes the term "indirect containment" (a holdover from I6, of course) to explain it. I think implying a different meaning to "containment" there than the one that applies in Inform muddies the water. But it *does* make sense to explain enclosure in terms of it being either direct or indirect... something. And today we have such a something, holding. It makes *both* of them easier to understand and to explain. And it at least reduces the number of exceptions that need to be noted to provide a full explanation of them.

And it's *useful* (as demonstrated by the Fallout Enclosure example, for one).

I expect the thing-requirement would break some existing programs, albeit a small number. But the chance of code breaking due just to changing the person requirement in the cases cited is much smaller still: I think it'd be reasonable to expect it to be nil.

I would sympathize with the rationale, at least, of changing its meaning to make it consistent with one or both of the other uses of `hold` as a verb. But it doesn't even have that to recommend it. I don't see what putative advantage restricting `if X holds Y` to things offers.

ganelson:
Yes, the containment, support, enclosure and incorporation relations will come next. Today we settle all relations business.

When regions contain each other, that can be tested with containment. If we have `Nirvana is a region. Shangri-La is a region. Nirvana is in Shangri-La.` then `Nirvana is contained by Shangri-La` and `Nirvana is in Shangri-La` are both true. But as you imply, that includes indirect containment. I don't think it's really problematic to omit this: "to hold" is a pretty odd choice of verb for this - would anyone naturally write `if Shangri-La holds Nirvana`?

If we allow `if X holds Y` to include non-things, while continuing to define the relation's domain as being (thing, thing), we go right back into the linguistic inconsistencies and logical errors pointed out in the original bug report. And I still don't really like saying that a room holds a thing. Hmm.

zedlopez:
> If we allow if X holds Y to include non-things, while continuing to define the relation's domain as being (thing, thing), we go right back into the linguistic inconsistencies and logical errors pointed out in the original bug report.

I'm suggesting that the relation's domain be (object, object) but that its usage in assertions and now-phrases be special-cased. In the proposal as written, assertion-hold and conditional-hold are *already* special-cased relative to now-phrase-hold. So this suggestion doesn't increase the amount of special-casing in the world, it just makes now-phrase-hold a special case instead of conditional-hold.

> And I still don't really like saying that a room holds a thing.

Inform will be continuing to say that via the `holder of` and `thing held by` phrases. (And "hold" has always been an even more awkward hypernym to embrace things incorporating things than it is in regard to rooms containing things.)

Excising one little bit of sub-optimal English usage in a single case of where a word applies such that the word has a separate meaning in *every* context in which it may appear constitutes a net loss for comprehensibility and usability.

curiousdannii:
Would this proposal turn holding into a generic (object tree based) relation any objects could use? If so that could be the solution for the spawning relation for Glk Windows (https://github.com/ganelson/inform-evolution/pull/12#issuecomment-1258909924).

ganelson:
I think I basically agree that (object, object) achieves better consistency, and experiment seems to suggest that this works well (at any rate, it breaks nothing in the test suite). Since this means embracing the idea that the verb is used more widely than just for things, I've made a more permissive `now X holds Y`, which distinguishes between objects used in the spatial world model from objects which are more abstract.

This may be a useful idea when we come to the harder problem of improving the consistency of `contains`.

ganelson:
To pose a question: given

https://inform7.atlassian.net/browse/I7-2219

should we alter HolderOf so that it returns `nothing` for direction objects? Given the current state of `to hold` in IE-0015, it's no longer true that `nothing holds north` (for example). But it's really very strange that direction objects seem to have a holder, and moreover, one with no natural description in Inform.

In an ideal world, the pseudo-object `Compass` would not exist, but it's not easy to get rid of. The parser depends on it, and has done since Inform 1.

zedlopez:
I favor HolderOf returning nothing for directions. It's hard to imagine that that change would break anyone's code. I had been going to bring it up, having noticed the inclusion of direction and region among spatial objects, but without further discussion. I'm guessing you were saving regions for discussion alongside containment. And I assume the spec will get explicit about how doors and backdrops will fit into the revised spatial relation scheme (beyond "awkwardly").

ganelson:
I've now made this change, and documented it in IE-0015. It caused no problems with the test suite. On further reflection, I'm sure this was a good change to make: it guarantees that `holder of` a spatial object is another spatial object. Previously, the `Compass` pseudo-object contradicted my claim yesterday that if `X holds Y` then either both are spatial objects or neither: because `Compass` is not a spatial object.

drpeterbatesuk:
Thanks Graham & Zed for this.

I see that there had already been a change between 9.3 and 10 such that 'Nirvana is in Shangri La' became true in 10.1, whereas it was false in 9.3 (where the compiler generally assumed regional containment as the meaning of 'in' or 'contains' in phrases like this, so conditions about one region containing another were always false unless the containment relation was forced, which could be done by the containing region being in an object variable (if \<obj-variable> contains <contained-region>) or the condition being expressed obliquely as '\<containing-region> relates to \<contained-region> by the containment relation'). Is this change documented somewhere? If not, is it worth documenting here?

The current proposal confuses me where it says that `something` means `some object` and therefore `something holds the player` is false. Surely a room is an object, holding the player? Shouldn't this say `something` means `some thing'?

Am I right in thinking that 'holds' now exactly corresponds to 'ParentOf()`?

How do you envisage authors reorganising the object tree of abstract objects at runtime if 'now X holds Y' is restricted to spatial objects?

drpeterbatesuk:
On the subject of spatial relations, is there a good reason why it's not possible to move rooms from one region to another at runtime (except through subterfuge, by 'manually setting' their map region property)?

Inform currently throws an error suggesting that moving a room to a region would imply that it would appear in all rooms in that region, like a backdrop, but plainly that isn't the case. If there are two regions, Indoors and Outdoors, then I board up the Verandah and set its map property to Indoors, it is simply now contained and regionally-contained by the Indoors region instead of the Outdoors. Does this cause a potential difficulty? And if not, why not allow 'now the Verandah is in Indoors?

drpeterbatesuk:
Another thing I've just noticed- since Ver 10, a region now appears in 'the list of objects to which it relates by the regional-containment relation' i.e. it appears to regionally-contain itself- which I assume is unintentional

drpeterbatesuk:
Also, for some reason the compiler's parser struggles with:

```if the region which relates to Verandah by the containment relation is Indoors: ```
but bracketing works:
```if (the region which relates to Verandah by the containment relation) is Indoors: ```
similarly,
```if Indoors is the region which relates to Verandah by the containment relation: ```
doesn't compile, but
```if Indoors is (the region which relates to Verandah by the containment relation): ```
does

**Comment by drpeterbatesuk** [comment]
Another change from 9.3 to 10 I've noticed: Regions now regionally-contain other regions that they enclose, which was not the case in 9.3.

In 9.3, regions contained, held and/or enclosed other regions but never regionally-contained them (despite what the docs said)

**Comment by drpeterbatesuk** [comment]
Ah. And, I see, that's the explanation for my first observation- the reason that 'Nirvana is in Shangri La' is now (since Ver. 10) true is _not_ because the compiler has changed its interpretation of the condition- it's still testing for regional-containment and not containment- it's just that regions can now regionally-contain each other (and even, it would seem, themselves!)

**Comment by drpeterbatesuk** [comment]
Ah. I see that the reason that 'Nirvana is in Shangri La' is now true is that since Ver. 10 regions can regionally-contain other regions (and, somewhat unexpectedly also regionally-contain themselves!) and _not_ because the compiler has changed its interpretation of the condition- it's still testing for regional-containment. It therefore remains the case that testing for _direct_ containment without the use of 'hold' is awkward.

**Comment by drpeterbatesuk** [comment]
Consequently, to test for *direct* containment of a region there is still no easy alternative to 'holds' or 'holder of'

**Comment by zedlopez** [comment]
Would something like the following be desirable?

```
Include (- Property spatial; -).
An object can be spatial or abstract. An object is usually abstract.
The spatial property translates into Inter as "spatial".
A direction is always spatial. A region is always spatial.
A room is always spatial. A thing is always spatial.
```

It seemed so to me when I first thought of it. On reflection, I realized there are good reasons to lump  rooms and doors together, or, more generally, rooms and things, or even rooms, things, and regions; the only time I ever want to lump all of the spatial objects together is when I'm testing I7 behavior. But I thought I'd toss it out in case someone else can think of why there would be a compelling reason for it.

**Comment by ganelson** [comment]
I was thinking about a "spatial" attribute just for efficiency reasons (testing X has spatial is quicker than X ofclass room or thing or region or direction), in fact.

Defining this at the I7 level as well is appealing in a way, but does occupy the nouns "spatial" and "abstract": we might want more obscure wording, and do we really want authors to be able to cause chaos by giving or removing this property at runtime?

**Comment by zedlopez** [comment]
> Defining this at the I7 level as well is appealing in a way, but does occupy the nouns "spatial" and "abstract": we might want more obscure wording

`representative` / `informative` ? Tough to thread the needle of obscure enough, but not too obscure. (In my own code, I've sometimes defined a `diegetic` adjective and created a `struct` kind of object for other things to be subkinds of... but I wouldn't inflict them on others.)

> do we really want authors to be able to cause chaos by giving or removing this property at runtime?

The latter doesn't trouble me so long as there's explicit documentation regarding "this is *possible*, but don't ever do it".

I've been assuming it's too radical a change to be plausible, but in case I'm wrong, I'll ask it: could the spatial objects' kinds be subkinds of a `game object` kind? It and `abstract object` would be the only two direct subkinds of `object` in this fever dream. I expect it would break the small number of extensions that have, say, K2_thing hard-coded, but I don't know whether it would also utterly break everything.

**Comment by ganelson** [comment]
Yes, I think that's a pretty feverish dream, as they go. I basically tremble at the thought of how many places in the code expect room and thing, etc., to be top-level kinds of object (in the compiler as well as the kits, not to mention who knows where in extension).

**Comment by ganelson** [comment]
So I have been looking at the regional containment relation this evening, prompted by Peter. Even this is not trivial. Firstly, I think it's quite right that regions should regionally-contain all regions they enclose: not doing so was just a bug in Inform 9.

But I think saying that a region regionally-contains itself is a bug in Inform 10.1, so I've fixed this.

Peter also notes that regional containment cannot be directly changed in 10.1. In fact, some slightly odd maneouvres could be done, but others couldn't. There's an argument against, but I've implemented `now X is in R` in cases where `X` is a room or another region, and this seems to work satisfactorily. See IE-0015 as revised.

Tonight's moral conundrum: Should `holder of M`, where `M` is a room, return the region which that room is in, rather than `nothing`? Equivalently: should a room be held by its (immediate) region?

**Comment by zedlopez** [comment]
> Tonight's moral conundrum: Should holder of M, where M is a room, return the region which that room is in, rather than nothing? Equivalently: should a room be held by its (immediate) region?

If we were spec-ing this out from scratch, I don't think the question would even come up: of course it should. Regions enclosing but not holding their rooms are one of the unfortunate anomalies in the current state of affairs.

As much as I'd like to see this change happen, there probably is code in the wild that could break in weird ways if the `holder of` a room ceased to be `nothing`. `while` loops advancing via `holder of` would be easy enough to spot, but it's easy to imagine it coming up much more subtly. And the effect, from an author's perspective could be the extremely frustrating "this thing that used to work now fails but *only in some rooms*."

All that said, it'd be *so* frustrating to *not* remove the inconsistency here that I lean toward yes, so long as a lantern is hung on the change, for instance a prominent notice in the IDEs. Maybe the first page of the docs, instead of its very neutral Change Log link could have a short message to the effect of: "10.2 (? 11?) cleans up some fundamentals regarding relations. It's possible, if unlikely, that stories written for previous versions may have depended on the previous messier behavior. See **changes since 10.1** to check whether it affects you."

**Comment by drpeterbatesuk** [comment]
>Tonight's moral conundrum: Should holder of M, where M is a room, return the region which that room is in, rather than nothing? Equivalently: should a room be held by its (immediate) region?

>All that said, it'd be so frustrating to not remove the inconsistency here that I lean toward yes

No, no, a thousand times no.  This is something that's not broken in any important way and does not need fixing, IMHO.

I can't see many circumstances where it would be really helpful. And it would break SO much code I suspect, both old and new, both I6 and I7, that assumes rooms are held by nothing-  even more so if this were to be implemented by the logical and pure method of doing away with the map region property and making rooms a child of their map region in the Object Tree.

I think the decision NOT to do this that was presumably taken when regions were first introduced to Inform was the only reasonable one at the time and much more so now, and the odd relations between rooms and regions that exist as a consequence should be regarded as an ancestral evolutionary quirk akin to the human hip joint- it's not how you'd design it from scratch, and it has a few weaknesses, but it basically works well and given its evolved interrelationships with other structures and mechanics we are where we are and it is is best not redesigned by overenthusiastic surgeons and their collaborators.*

*50 years of experience in the joint replacement field has proven this to be true

**Comment by ganelson** [comment]
Well, always good to reach a consensus.

Enclosure is defined at runtime by the function `IndirectlyContains`. It's not quite the transitive closure of holding, but close. There seem to be two carved-out exceptions: (i) a room encloses a door if either side of the door abuts it, and (ii) a backdrop is never enclosed by anything. I'm dubious of (ii), but backdrops are very difficult to make sense of in any of the spatial relations. It would be possible to do something parallel to (i): i.e. a room encloses all the backdrops present in it. But perhaps the reason we currently say no is that backdrops cannot be fully enclosed by a room. If a river flows through three rooms, would we say that none of the three enclose it, or that all of them do? Some of this gets to the awkward point that the name of the function, "indirectly contains", is perhaps a different idea than the name of the relation, "encloses".

Currently the code looks like this:

```
[ IndirectlyContains o1 o2;
	if ((o1 == nothing) || (o2 == nothing)) rfalse;
	if ((o1 ofclass K1_room) && (o2 ofclass K4_door)) {
		if (o1 == FrontSideOfDoor(o2)) rtrue;
		if (o1 == BackSideOfDoor(o2)) rtrue;
		rfalse;
	}
	if (o2 ofclass K7_backdrop) rfalse;
	for (o2 = HolderOf(o2) : o2 && o2 ~= thedark : o2 = HolderOf(o2)) if (o2 == o1) rtrue;
	rfalse;
];
```

It seems to me that the `o2 ~= thedark` clause is redundant? Surely no object is ever a child of the `thedark` pseudo-object. Am I wrong about that?

That leaves `for (o2 = HolderOf(o2) : o2 : o2 = HolderOf(o2)) if (o2 == o1) rtrue;` which means that, with the exceptions cleared out of the way, enclosure is the transitive closure of holding. This has a pleasing clarity.

This I think is the real significance of the issue of whether a region can be the holder of a room. At present, a region can only enclose another region. Surely a region does enclose a room which is part of it, and therefore also anything in that room? (I note that, at present, a region _is_ the `container of` a room, but not the `holder of` that room.)

So today I have two questions:

(a) Should we make a room enclose any backdrop found in it?

(b) On the subject of regions enclosing stuff, which is the more attractive option?

* Making enclose the transitive closure not of `HolderOf` but of some modified function, `RegionFixedHolderOf`, say, in which a room does return its region.
* Make the holder of a room its region after all.
* Declare that a region doesn't enclose anything.

**Comment by drpeterbatesuk** [comment]
>Surely no object is ever a child of the thedark pseudo-object. Am I wrong about that?

It's an awkward thing to arrange in I7, and I doubt any existing code makes use of this remote possibility- particularly since I7 has already deviated from I6 in that in I7 when in darkness children of `the dark` remain off-stage and out of scope.
>(a) Should we make a room enclose any backdrop found in it?

Do you mean all rooms which would be returned by inspecting the backdrop's I6 found_in property would enclose it? or just the room (if any) the backdrop is currently a child of in the Object Tree?
If the former, my initial thought is that this would mainly be helpful as a means to knowing if a backdrop *would* be moved to a location if the player were there. Currently, I think this requires either making a call to BackdropLocation() in I6 or perhaps: surreptitiously move the player to the room in question, update backdrop positions, make the check, then surreptitiously move player back to their starting location. The potential downside to implementing this particular solution to that gap in I7 is that loops in existing code that iterate over things enclosed by a room, which are probably quite common, might begin to have unexpected results, or cause runtime errors due to attempted illicit manipulation of backdrop objects.

I would be in favour of an inbuilt I7 verb- 'found in' is probably as good as any- to mean where doors and backdrops would appear if the player were to be that location, but without making rooms enclose backdrops.

Making just the room currently holding a backdrop also enclose it would tidy up the nice 'enclosure'=='(indirectly) hold', 'hold==parent of' paradigm, but it may be best to abide by the principle of 'if it ain't broke don't fix it' and leave things as they are.

>(b) On the subject of regions enclosing stuff, which is the more attractive option?
>
>- Making enclose the transitive closure not of HolderOf but of some modified function, RegionFixedHolderOf, say, in which a room does return its region.
>- Make the holder of a room its region after all.
>- Declare that a region doesn't enclose anything.

There already exists an indirect relation that transcends the region/room barrier- regional containment- and which now encompasses all of regions, backdrops, rooms and their contents, direct and indirect.  I'm not sure much is to be gained by extending enclosure across the region/room barrier.

So I would lean towards leaving things as they are, with the 'cascade of enclosure' broken at the region/room interface, with regional containment being the 'one relation to rule them all'.

It's already straightforward enough I think to access the 'lowest in the hierarchy of regions' region of a room, or of any of its contents, via `the map region of <a-room>` or `the map region of the location of <an-object>`.

>If a river flows through three rooms, would we say that none of the three enclose it, or that all of them do?

The same concern already applies to backdrops and regions. I think the present situation is that if a backdrop is 'found in' any of the rooms regionally contained by a region, the backdrop is said in conditions to be 'regionally in' said region, whether its current position in the Object Tree reflects that or not.

**Comment by drpeterbatesuk** [comment]
PS while we're thinking about regional containment and backdrops, I filed a bug report a while back about how assertions regarding backdrops are broken in Ver 10- it's no longer possible to mix `<a-backdrop> is in <a-region>` with one or more `<a-backdrop> is in <a-room>` : the former prevents any of the latter being implemented.

**Comment by drpeterbatesuk** [comment]
PPS I don't think it hurts to keep enclosure of regions by regions. It would break some existing code to remove it and its presence keeps enclosure almost the same as indirect holding.

I think if you want to have a relation as the transitive closure of RegionFixedHolderOf(), that can be seamlessly applied both above and below the region/room barrier, perhaps call the direct/indirect relations something new, different from hold/enclose. encompass/regionally encompass or simply regionally hold/regionally enclose? We're running out of synonyms here!

**Comment by zedlopez** [comment]
Just to track things, the backdrop issues being touched on:
- [I7-2215](https://inform7.atlassian.net/browse/I7-2215): backdrops that are associated with a specific room or list of rooms are held by nothing until the player has visited
- [I7-2254](https://inform7.atlassian.net/browse/I7-2254): the location of a backdrop is poorly-defined, with the knock-on effect that 'if a-backdrop is nowhere' is poorly-defined

Peter's right, rooms should remain held by nothing. At some level I knew I needed someone to hold me back; that's why I went on about potential to break code.

> (a) Should we make a room enclose any backdrop found in it?
>
>(b) On the subject of regions enclosing stuff, which is the more attractive option?
>
>    Making enclose the transitive closure not of HolderOf but of some modified function, RegionFixedHolderOf, say, in which a room does return its region.
>    Make the holder of a room its region after all.
>    Declare that a region doesn't enclose anything.

And again, in the interests of backwards compatibility, the behavior should probably be left as is. And regions enclosing backdrops would probably make things *more* confusing since several discontinuous regions might regionally contain a given backdrop (since it could be defined as in multiple rooms and those rooms could be in different regions) and it'd be easy for an author to assume that for a region enclosure would be a superset of regionally-contained and it's probably easier for regions not to enclose non-regions than it is to remember the difference.

Something that wouldn't break anything, though, would be new relations that behave how hold and enclose would if they were consistent. Not technically difficult, of course; the hardest problem would be coming up with good names. The worst side effects would be overhead the Z-machine crowd might not want and potential to confuse new users more by having things so much like hold and enclose. Untried code that should convey the idea:

```
To decide what object is the subsumer of (o - an object):
  if o is a room, decide on the map region of o;
  if o is a backdrop, decide on the location of o;
  decide on the holder of o;

Subsumption relates an object (called the parent) to an object (called the child) when the parent is the subsumer of the child.
The verb to subsume means the subsumption relation.

To decide if (o1 - object) indirectly-subsumes (o2 - object):
  let obj be the subsumer of o2;
  while obj is not nothing begin;
    if obj is o1, yes;
    now obj is the subsumer of obj;
  end while;
  no.

Envelopment relates an object (called the ancestor) to an object (called the child) when the ancestor indirectly-subsumes the child.
The verb to envelop means the envelopment relation.
```

The above would inherit the weirdness surrounding doors and holding.

Edited to add: since this is new, I can be a little more radical, and what I had above didn't work well for backdrops, so: unlike with holding, a thing can be subsumed by more than one thing, so doors are subsumed by the rooms on both sides. 

```
To decide if a/an/-- (B - a backdrop) would be observable in a/an/-- (R - a room): (- (BackdropLocation( {B}, {R} )) -);

To decide if (o1 - object) directly-subsumes (o2 - object):
  if o2 is nothing, no;
  if o2 is a room and the map region of o2 is o1, yes;
  if o2 is a door and o1 is a room and o1 encloses o2, yes;
  if o2 is a backdrop and o1 is a room and o2 would be observable in o1, yes;
  [ edited to remove wrong-minded complication about regions and backdrops. In this picture, a region wouldn't directly subsume anything but a region or a room ]
  decide on whether or not o1 is the holder of o2;

Subsumption relates an object (called o1) to an object (called o2) when o1 directly-subsumes o2.
```

**Comment by drpeterbatesuk** [comment]
I've been giving this some more thought, that thought being that if we were to make the leap to saying that the map region of a room encloses but does not hold it (with precedent the room on the 'other' side of a door encloses but does not hold it) and that consistent with that one anomaly, a region encloses everything it currently regionally contains, and that a region also encloses the backdrops it currently regionally contains, we'd pretty much make regional containment redundant- which would be a path to pare back to the number of spatial relations and synonyms required, rather than proliferate them.

I doubt very much code is reliant on the fact that regions currently don't enclose rooms and their contents, so I doubt much would break. Regional containment could perhaps be flagged for future phasing out.

An alternative would be keep enclosure as it is and to say that regional containment can now apply not just to regions regionally containing regions, rooms, backdrops and things but is extended to rooms regionally containing things or backdrops (the latter being true if a backdrop is 'found in' that room) and to backdrops and things regionally containing things (in the case of backdrops, things they directly or indirectly incorporate). This sounds weird, but almost certainly wouldn't break anything. It could be given an alternative name to help authors queasy about saying that a jam jar regionally contains frogspawn, but really there would be no need because they could continue to use 'enclose'.

`The regional container of` (or some renamed version) would be the counterpart of `the holder of` for this all-encompassing relation.

Worth considering, perhaps.

PS This is the backdrop issue I was alluding to: [I7-2227](https://inform7.atlassian.net/browse/I7-2227) In 10.1.2, only one of multiple statements to place a backdrop is implemented when at least one is implemented as a routine

**Comment by ganelson** [comment]
We'll get to enclosure, eventually, but I've taken a step backwards first and specified the relations tackled so far more carefully. I hope the proposal as rewritten is more explicit. In the process of doing this, I found a couple more minor edge cases in holding and regional containment, which are now fixed. The reason for the strange treatment of `thedark` eventually came to light: it wasn't at all what I thought (i.e. vestigial code) but was a clumsy bug-fix of an issue to do with the state of the world before the player's initial situation is set up at runtime.

For the moment, at least, regions continue not to hold rooms. I've thrown in the towel on hoping that regional containment might be the transitive closure of holding: instead it has a weaker transitive property, i.e., that if R contains X and X holds Y then R contains Y as well. Even that required a little work to extract, but the process should fix some weird potential bugs to do with component parts of backdrops and doors.

I've also added "support" to the set of relations tackled. This needed a little more defence than it was previously getting, but is otherwise okay.

Today's quiz questions:

* Should we allow `now P is not wearing X`? This would do nothing unless `P` is wearing `X`, and in that case, change the state so that now `P is carrying X`.
* Should we allow `now P is not carrying X`? This would do nothing unless `P` is carrying `X`, and in that case, move `X` to the holder of `P`.

**Comment by drpeterbatesuk** [comment]
**Regarding the Specification**

>Note that nothing means `no object`, not `no thing`. Since a room, in this case Lab is an object, it is not true that nothing holds the player. On the other hand, `something` means `some object`. So despite appearances, it is perfectly consistent that nothing holds the player and something holds the player are both false.

I assume that this should read `something` means `some thing`.

As an oblique observation, it's one of the more recurring themes on the intfiction forum that it's confusing that in different contexts 'thing' and 'something' mean different things- also 'visible'

For example, in action definitions such as `nit-picking about is an action applying to one visible thing` `visible thing` means `object existing in the story universe`

But in grammar tokens:
 `[thing]` means `thing`
 `[visible thing]` means `thing which can be seen by the player`
 `[something]` means `object`
 `[object]` isn't allowed (and directs the author to use `thing`!)
 `[things]` means `objects`
 `[objects]` means `objects`
 `[any thing]` means `any thing`
 `[any something]` means `any thing`
 
Yet in action descriptions, such as `Carry out nit-picking about something` `something`/`anything`/`thing`/`some thing`/`a thing` are all synonymous in restricting the noun to things, (any thing is not allowed here)- but you can if you wish also say `Carry out nit-picking about object` (or `some object` or `an object`)

Although a source of complaint, I suspect the mixed-up semantics of grammar tokens is far too deeply embedded to change without unleashing Armageddon on the existing code base.

Whether there is merit in allowing forms of expression for action definitions more consistent with usage elsewhere, such as `nit-picking is an action applying to one in-world object` and to promote these as preferred usage is perhaps worth considering.  It's also been raised on the forum that not only can a 'visible thing' be neither visible or a thing, a 'touchable thing' can also be neither touchable or a thing, as the basic accessibility rule is not applied to non-things such as rooms and regions (https://intfiction.org/t/a-touchable-thing-can-be-neither/56131/9)

>Asserting that `now X holds Y` imposes a new restriction if either `X` or `Y` is a spatial object. In these cases, `Y` must be a thing, and `X` must be either a room, a person, a container or a supporter.

I think it might be reasonable, for consistency of expression, if this were to be extended to allow `now X holds Y` when `X` and `Y` are both regions. It's difficult at present to talk about direct containment insofar as regions are concerned because the compiler wants to assume regional containment is implied by conditions `if X is in Y` or 'if X contains Y` in contexts when `X` will always oncontrovertibly be a region. The only easy way to talk about direct containment for regions is to use 'hold'.

>- If both are abstract objects, then now X holds Y is permitted in all cases except that in which Y already directly or indirectly holds X.

I don't understand the restriction here. Is not this the equivalent (for spatial objects) of taking a carrier bag out of a box & putting the box in the bag? Why should that be disallowed?

>If the Brass Door is between the Blue Room and the Green Room, then only one of the Blue Room holds the Brass Door and the Green Room holds the Brass Door will be true, and it will not always be easy to be sure which.

Which,I presume, is the perceived utility of `if the Blue Room encloses the Brass Door` being (anomalously) true either way, avoiding as it does, at the price of consistency, having to say `if the Blue Room holds the Brass Door or the Blue Room is the other side of the Brass door` or similar.

>There is a tenuous argument for this, but it seems more likely to surprise authors than not to. (Is Europe in Europe? I think I would say not.)

And probably a strong argument against- nowhere else in Inform can an object be said to be `in` itself. Outside that consideration, it's a bit like the old 'Should a library catalogue contain itself?' paradox.

>in the case where X equals or is contained by R, a runtime problem is thrown:

Same observation concerning carrier bags and boxes. I think I must be missing something here. Although it's more difficult to think of obvious examples of where this would be needed, I don't see why logically it should be precluded.

>No verb directly means regional containment, but assertions such as `X is in R`. can be made for regions or rooms `X`, and these are then read as using regional rather than regular containment.

Although, if `X` is a room, it will as an end result necessarily be both contained and regionally contained by `R`. Furthermore, since the assertion `X is in R` can only assert direct containment between regions, regular containment is also necessarily implied by every such assertion. Unless we stop saying that regions regularly contain regions that they hold.

>`Y` is a two-sided door which is not `absent`, either of whose locations is such a room.

That's interesting. I didn't know a door could be `absent`. How is that achieved (in I7) and how is it manifest? Or is this something new?

In terms of how this and the previous specification are phrased, a backdrop does not have any well-defined location and a door has only one at a given time, so this would perhaps be better expressed in terms of the room(s) they are found in?

>now Y is [regionally] in X is permitted only when Y is also a region.

As suggested above ,it might perhaps be better to frame `now Y is in X` as always implying direct regular containment, with the creation of regional containment as a potential side-effect if `X` is a region or already regionally contained by one or more regions. Either way, Y can be a region *or* a room.

**Regarding Today's Quiz**

>-Should we allow `now P is not wearing X`? This would do nothing unless P is wearing X, and in that case, change the state so that now P is carrying X.
>-Should we allow `now P is not carrying X`? This would do nothing unless P is carrying X, and in that case, move X to the holder of P.

I think these would be nice conveniences to the author, and certainly read more naturally than, for example `if Professor Yaffle carries the Collected Works Of Proust, move the Collected Works Of Proust to the holder of Professor Yaffle`

**Comment by ganelson** [comment]
I have rephrased:

Using `now` to change holding can result in a situation where an object
indirectly or directly holds itself. This is not allowed, either for spatial
or abstract objects, but technically results in a "programming error" message
rather than a run-time problem message.

...but no, you definitely can't safely execute:

```
now XX holds YY;
now YY holds XX;
```

The result of the second phrase here is not analogous to taking a box out of a bag and then putting a bag into a box. It would be putting the bag into the box while the box is still in the bag.

I agree, `now R holds S` should be allowed in the case where both are regions, and I've made this change.

I'm not at all sure whether a door can usefully be `absent`, but the code in `WorldModelKit` appears to anticipate that contingency. For now, that's a side-issue, though.

**Comment by zedlopez** [comment]
I will catch up on everything later, but thought I'd quickly note (and continuing the theme of Dr. Bates and I ending up mutually contrarian):

> -Should we allow now P is not wearing X? This would do nothing unless P is wearing X, and in that case, change the state so that now P is carrying X.
> -Should we allow now P is not carrying X? This would do nothing unless P is carrying X, and in that case, move X to the holder of P.

I feel iffy about this.

In the general case for relations, there aren't side effects. `now X does not V-meaning-R Y` is a no-op if X didn't already V Y; if X did V Y, that is no longer true, without other consequence (beyond that  in a symmetric relation it means Y will no longer V X either). 

Like the familiar error message for "encloses" and some cases for "holds" says: "this is not something I can make true with 'now', because it is too vague about the underlying cause which would need to be arranged." If X carries Y, there are a zillion ways it could be the case that X does not carry Y. (The wears case is more defensible: if X wears Y, we know X possesses Y, so X not wearing Y implies removing the worn attribute discretely.)

 If R1, R2, R3 are regions and R1 holds R2 and R2 holds R3, we wouldn't expect that `now R2 does not hold R3` would result in R1 holding R3.

I think it might be clearer to consistently keep it the case that `X does not hold Y` can only be used when it implies a discrete change in game state without other side effects. It might be a feature to make authors spell out `move Y to the holder of X` because that leaves no opportunity to be surprised by the effect: what happens is exactly what they said should happen.

(I'll admit, something I've craved in this world is hooks on establishing and dissolving relations: something like a rulebook for each case based on a parameter object with three properties: the relation, X, and Y so you *could* have side effects and plenty of 'em... but I massively digress.)

**Comment by ganelson** [comment]
I certainly don't think `now X does not hold Y` should be allowed, and perhaps you're right that `does not carry` is a similar situation. But `now P is not wearing X` - well, I think it's clear how to make that true. Shall we do just that one?

**Comment by ganelson** [comment]
I have now added details of two more spatial relations: room-containment (which I'll hazard a guess nobody but me even knows the existence of), and enclosure. Both required minor consistency tweaks.

It really is unfortunate how two-sided doors and backdrops are handled differently in these different relations. For holding, both are in the place most recently seen by the player. For room-containment, both are in either the player's current location or, if that is not an option, in the first room they belong to. For enclosure, doors are in both possible locations, but backdrops are in none of the possible locations. For regional-containment, doors are in both possible locations, and backdrops are in all possible locations, but only if they are not `absent`.

Are we comfortable with this? If not, can we do anything to simplify matters?

**Comment by drpeterbatesuk** [comment]
>which I'll hazard a guess nobody but me even knows the existence of

Ha! You guess mistakenly! (excerpt from [my interminably long post on intfiction forum re spatial relations](https://intfiction.org/t/elegant-way-to-test-whether-something-is-in-a-location-even-if-in-a-container/49420/7?u=drpeterbatesuk), now up to a startling 62 edits- with more to come, it would seem,)

>In addition to the direct basic spatial relationships there are some indirect spatial relationships:
>
>-enclosure applicable to a room or thing directly or indirectly holding a thing or to a region directly or indirectly holding a region
>-room-containment applicable to a room directly or indirectly holding a thing
>-regional-containment applicable to a region directly or indirectly regionally-containing a room, or a thing enclosed by that room, or a backdrop [and since Ver.10 a region]
>
>Because the first two of these indirect spatial relationships never specify exactly where and how the contents are held, they can’t be used in assertions. You can’t for example write ‘Now John encloses the teapot’ or ‘John encloses the teapot’.
>
>Room-containment also has no specific built-in verb. It usually appears in source as conditions like ‘if the Kitchen is the location of the teapot’. You can define a verb if you like: ‘The verb to locate means the room-containment relation. The verb to be located in means the reversed room-containment relation’, after which you can write the equivalent condition ‘if the Kitchen locates the teapot’ or ‘if the teapot is located in the Kitchen’ although ‘if the Kitchen encloses the teapot’ or ‘if the teapot is enclosed by the Kitchen’ would do just as well. Note that ‘if the teapot is in the Kitchen’ always implies direct containment, not room-containment, so you must use ‘the location of’ or ‘encloses’ or a specifically-defined verb like ‘locates’ if you want to include things indirectly contained or held by a room.

-----------------------------------------------------------------------------------------------------------------------------------------

>For room-containment, both are in either the player's current location or, if that is not an option, in the first room they belong to.

This is a problem particularly for backdrops, because it leads to the answer to `if <a backdrop> is nowhere` being poorly-defined.  See discussion [here](https://intfiction.org/t/elegant-way-to-test-whether-something-is-in-a-location-even-if-in-a-container/49420/26)

My gut instinct is to implement what the WI docs say- that backdrops have NO location (i.e. the location of a backdrop is always nothing). This seems in one sense naturally true, as they are not confined to one location. I don't think this is necessarily inconsistent with a different definition for the location of doors, which are fixed in place BETWEEN locations. However, it would not be unreasonable to say that doors also always have no location, being always BETWEEN two locations (or between one location and nowhere) and within neither.

I'm not sure this would break much existing code, since 'the location of' both doors and backdrops has always been difficult to keep track of and not of much practical use to know.  It's much more useful to know whether a backdrop 'is found' in a room than where it actually is right now. `if <a door> is in the location` is the natural way to express the condition of the door being in the same room as the player and `if <a room> encloses <a door>` the natural way to express the condition of a door being 'in' (in a natural sense, not a World Model sense) a room. We can still find out which room currently holds a door, if needed, through the holding or containment relations.

>I certainly don't think `now X does not hold Y` should be allowed, and perhaps you're right that `does not carry` is a similar situation. But `now P is not wearing X` - well, I think it's clear how to make that true. Shall we do just that one?

I'd still vote in favour of `now x does not carry y`, although the same thought had occurred to me regarding 'there are lots of ways to arrange that, and you can't expect me to choose one'. Perhaps the same idea  of  `now <a-person> forcibly and silently drops <a-thing> if they currently carry it (or perhaps, enclosed by something carried) (without invoking the dropping action)` could be phrased differently and added to the language, to save Zed's queasiness? I have to say, the intended meaning and expected result of `now x does not carry y` is unlikely to be in doubt to whoever wrote it, despite acknowledging Zed's more purist and logical position.

>...but no, you definitely can't safely execute:
>
>`now XX holds YY;`
>`now YY holds XX;`
>The result of the second phrase here is not analogous to taking a box out of a bag and then putting a bag into a box. It would be putting the bag into the box while the box is still in the bag.

I can see what you mean- again I was approaching this from the woolier 'natural language idiom', in which it would be obvious to a person instructed to put the bag into the box that removing the box first is required. Although not to the Inform `inserting it into` action, I notice, which does reject such an idea at the Check stage with the `can't insert something into itself rule`. I would contend that this rule should be smarter about the situation :-) . It's more obvious as assertions while arranging the initial state of the world that both these phrases can't be true _at the same time_. The more unnatural and longwinded alternative (without invoking the controversial `now XX does not hold Y` is, I guess:

`now XX holds YY;`
`now YY is nowhere;`
`now YY holds XX;`

I guess my thinking was that the intermediate state of YY being nowhere is sort of implicit in any `now XX holds YY` phrase.

**Comment by drpeterbatesuk** [comment]
>For enclosure, doors are in both possible locations, but backdrops are in none of the possible locations.

My take on this is that enclosure for backdrops should be made consistent with doors- i.e. all rooms a backdrop is found in enclose it- which is not only also consistent with the treatment of regional containment but would provide the (currently missing) means to know whether a backdrop is found in a specified room. Having nothing enclose a backdrop doesn't help an author in any way that I can think of.

**Comment by drpeterbatesuk** [comment]
>now `X is roomwise in Y` is permitted only if `X` is a thing and `Y` is a room, and is then equivalent to `now X is in Y`.

This is another of those 'Too many ways to arrange it' things. Again, I'd be in favour (despite anticipating purist objections) in that the intent of the author and how to arrange that outcome is obvious. That said, despite your reservations, the same could be said of `now X holds Y`- the intent of the author is clearly in I6 terms `move Y to X`, such that `child(X)==Y` is true; equivalent in idea to the currently non-permitted I7 phrase `now the holder of Y is X`. It merely requires type-checking to ensure `X` is one of an abstract object, a region, a room, a person, a container or a supporter and that: if `X` is a region, Y is also a region; if `X` is an abstract object, `Y` is also an abstract object; otherwise, that `Y` is a thing that is not a door or a backdrop.

>in 10.1, a room room-contains itself. So for example if Chapter House is a room, then Chapter House is somewhere is true, and Chapter House is nowhere is false;

This isn't something I ever realised.  Do we want this? Analogous to regions not regionally containing themselves, I would have thought that rooms shouldn't room contain themselves. The more important question is probably `Is it useful to authors to say that a room has a location, which is itself?' or that 'a room is always somewhere, never nowhere and sometimes everywhere'.  I'm struggling to see the usefulness of that. And if it's not useful, I'd say that the location of a room is nothing, meaning that it is also always nowhere, never somewhere and never everywhere. Having said that, I really don't have strong feelings about it either way- and clearly at one point others felt differently.

**Comment by ganelson** [comment]
In the interests of making progress by removing one topic of debate here, I have enabled `now X does not wear Y` (clearly harmless) and `now X does not carry Y` (the only objections to which are fairly theological). I think these are of some potential use and do no harm. Carrying and wearing are now my absolute favourite relations, because they DO NOT INVOLVE REGIONS, BACKDROPS OR DOORS.

**Comment by drpeterbatesuk** [comment]
>Carrying and wearing are now my absolute favourite relations, because they DO NOT INVOLVE REGIONS, BACKDROPS OR DOORS.

lol :-)

**Comment by ganelson** [comment]
I've broken ground on containment, which is a mess. I think I've improved its worst aspect, but not in a completely painless way. For the full tale of woe, see IE-0015.

**Comment by drpeterbatesuk** [comment]
**From the specification**

>(3) The third problem arises from the way that `X is in Y` is interpreted as meaning containment most of the time, but as regional containment when Y is the literal name of a region.

Actually, I think the compiler is smarter than this. It interprets `if X is in Y` as regional containment whenever it can be determined at compile-time that Y is a region- which includes not only when Y is a literal name of a region, but also when Y is a local or global region variable.

I think an improvement for (3) would be to introduce more elegant phrasing than is currently available out of the box to force testing for regular or regional containment. For the latter, `if X is regionally in Y` / `if Y regionally contains X` makes sense (I think at present the compiler just ignores the `regionally` if you write this- the exception being if X is a backdrop, in which case the 'regionally' is *required*). For the former one can always use `hold`, but `directly contains` / `directly in` would be a more natural usage and would also mean containment rather than any other variety of holding.

**Comment by ganelson** [comment]
I have investigated and added visibility, audibility, touchability, and concealment. The three -bilities needed very small changes to ensure they are for things only. This gave the test suite no concerns, and I think it's unlikely to cause problems.

What was interesting was to find that touchability goes to elaborate trouble to get the case of backdrops/doors very nearly right. The rest... do not.

**Comment by drpeterbatesuk** [comment]
Infelicities of scope and visibility:

```
"Scratchpad_Visibility" by PB

The the Lab is a room. The the cellar is a dark room. The cupboard is an open enterable container in the cellar. Bob is a man in the cupboard. The diamond is in the cellar. The retort is in the Lab. Persuasion: rule succeeds. The the dispensary is a dark room. It is north of the Lab. Sue is a woman in the Dispensary. The morphine is in the Dispensary. The indefinite article is "some". The block giving rule does nothing. The player carries a transparent closed container called a jam jar. The frogspawn is in the jam jar. The description is "Like one hundred eyes looking back at you.".The player carries an opaque closed container called a matchbox. The spider is in the matchbox. The description is "A monstrosity comprising too many hairy legs and a dozen reflective eyes."
After deciding the scope of the player: place Bob in scope.
After deciding the scope of the player when the location is the location of Sue: place Sue in scope.
When play begins: describe scope and visibility.
Every turn: describe scope and visibility.
To describe scope and visibility:
	repeat with p running through people:
		say "[line break][bold type]For [p]:[roman type][line break]";
		repeat with o running through things:
			say "[the o] is [unless o is in scope for p][bold type]not [roman type][end if]in scope and [unless o can be seen by p][bold type]not [roman type][end if]visible.";
	
Test me with "Bob, pick up the diamond/Bob, examine the retort/n/take morphine/Sue, give me morphine/i/x frogspawn/smell frogspawn/x spider"
		
To decide whether (o - an object) is in scope for (p - an object): (- TestScope({o},{p}) -).
```

**interesting observations**
(i)  an NPC placed in scope for an actor also becomes visible to the actor, but only while the actor is illuminated.
(ii)  the actor can observe the placed-in-scope NPC's actions even though they take place in another room and in the dark.
(iii) darkness does not affect what is in scope for NPCs, meaning NPCs can perform actions in the dark which are impossible for the player.
(iv) in darkness, things enclosed by the player in transparent closed containers remain in scope- presumably in darkness there should be no distinction between an opaque and a transparent container. That said, this quirk has less impact on play than it might at first appear, since there are no built-in actions that act in darkness on inaccessible things. The refusal message just becomes either 'It is pitch dark and you can't see a thing.' (... is an action...one visible thing...and requiring light)-failing basic visibility rule, or '<the container> isn't open'(... is an action...one touchable thing.)-failing basic accessibility rule, instead of 'You can't see any such thing.'-parser 'out of scope' error.
(v) attempts to persuade an NPC to perform an action on something out of scope for them are converted to an 'Answering \<the NPC> that \<topic>' action, which requires accessibility and, if the NPC is inaccessible, a confusing 'you can't reach into \<the location of the NPC>' response.

**Comment by drpeterbatesuk** [comment]
**I fear that concealment has multiple issues in 10.1**

I'm uncertain what the specification for concealment is or was, but I have tended to assume it is to allow things to hide things that they hold as if, to an actor, they were not there (the equivalent, perhaps of them being moved temporarily to 'nowhere' without the author having to perform that time-honoured sleight of hand)

They should therefore be:

- Out of scope to the actor
- Not able to be seen by the actor
- Not revealed when examining or searching or looking under or opening the concealing object
- Not revealed in any part of locale descriptions
- Excluded from 'all' and similar commands
- Excluded from inventory (debatable)
- Never inferred
- Never included in disambiguation questions

Against this standard at least there are a number of issues with concealment as currently implemented.

(i) if one or more concealed things are the only things on a scenery supporter,
	a) the describe what's on scenery supporters in room descriptions rule is not suppressed as it should be
	b) the presence of at least one concealed item is revealed to the 'in-the-know' by the locale description 'on the scenery supporter is nothing'
This unfortunate situation is due to concealed things not being included in the definition of locale-supportable things, which does exclude scenery and undescribed things.
(ii) similar problems arise when examining containers or supporters, or opening containers- the list-writer is invoked but then prints the text 'nothing'- a different response to containers or supporters that are truly empty.
(iii) the direct possessions of an actor cannot be concealed from the actor. There is a certain sense to this idea, but authors do sometimes wish to conceal possessions from the player and prevent them being dropped etc. in which instances this is a frustrating restriction. The agency here should perhaps be with the author, not the World Model.
(iv) concealed things are listed in inventory even when not direct possessions of the player. Again, one can make an argument for this, but it means that if a player is able to take possession of a concealing thing, its concealed contents can be revealed by taking inventory, which is not likely to be what the author intended.
(v) rooms can't conceal things, only things can conceal things. Again this seems an unnecessary restriction on the author. In particular, it means that doors can't be concealed. They can be undescribed and scenery, but that doesn't remove them from scope and allows them to potentially be included in disambiguation questions- in particular if a room contains more than one such door which therefore score equally in ScoreMatchL() : e.g.

>\>open door
>  Which do you mean, the Oak Door or the Iron Door?

(vi) The definition of concealed as an adjective requires the concealing thing to be the holder of the concealed thing. However, the concealment relation allows things enclosed by the concealing thing to be concealed by it. This means that 'something conceals X' can be true, while 'X is concealed' is false. (This is an improvement on, or at least a change from, 9.3, when a concealing thing could conceal any thing anywhere.) Furthermore, when something is concealed by a concealing thing (relation)but not concealed (adjective), in play concealment does not apply- i.e. it is in scope and visible etc.
e.g.
>The Lab is a room.The vanishing box is a container in the Lab. In the vanishing box is the magic hat. In the magic hat is the rabbit.
>
>Rule for deciding the concealed possessions of the vanishing box when the particular possession is the rabbit: decide yes.
>
>When play begins:
>	say "The vanishing box does [unless the vanishing box conceals the rabbit]not [end if]conceal the rabbit.";
>	say "The rabbit is [unless rabbit is concealed]not [end if]concealed.";

leads to

>The vanishing box does conceal the rabbit.
>The rabbit is not concealed.
>
>**Scratchpad_Visibility**
>An Interactive Fiction by PB
>Release 1 / Serial number 230127 / Inform 7 v10.1.2 / D
>
>**Lab**
>You can see a vanishing box (in which is a magic hat (in which is a rabbit)) here.
>
>\>take rabbit
>Taken.

**Comment by zedlopez** [comment]
I wrote fixes for Peter's cited (i) and (ii) concealment issues above and had been planning to submit a PR after more testing. At the moment, I'm without a working compiler but hope to remedy that this weekend.

My guess had been that the behavior in (iii) through (v) would be considered features not bugs and, unusually, I don't have strong opinions on the subject. Well, except that there's something lurking in (iv) that I consider distinctly a bug: if the player carries a supporter or container that conceals something it holds, the "concealed" thing is listed in inventory yet is out of scope. Either the contents should be mentioned *and* be in scope or neither mentioned nor in scope.

There are some creaky old Mantis bugs touching on (i) and (ii).

[I7-2046](https://inform7.atlassian.net/browse/I7-2046) [Mantis 2083] Containers holding concealed items do not show the " (empty)" detail text
[I7-2036](https://inform7.atlassian.net/browse/I7-2036) [Mantis 2073] Both examine containers and examine supporters rules inconsistent when contents are concealed

Speaking of scope, here's some scope creep: if there are concealment rules, they get checked a zillion times over during each trip through the parser. I have felt it would be a big improvement if most of the behavior in [Scope Caching by Mike Ciul](https://github.com/i7/extensions/blob/9.3/Mike%20Ciul/Scope%20Caching.i7x) were built-in.

(My not meaningfully tested fixes are in [Concealer](https://github.com/zedlopez/extensions/blob/concealer/Zed%20Lopez/Concealer.i7x). I had forgotten that it does make a call on (iv): inventory doesn't print concealed contents of containers or supporters.)

Edited to add: here's a borogove snippet demonstrating the issues discussed in (i), (ii), (iv). https://snippets.borogove.app/inform7/z2qsv7

And here's the same setup with an improved version of the fixes from Concealer: https://snippets.borogove.app/inform7/b9dnbr

**Comment by ganelson** [comment]
The concept of concealment is that whatever is doing the concealment is hiding something from the world outside. For containers, that seems clear enough. For the case of a person concealing something, the person knows perfectly well that the concealed thing exists: the person is merely hiding it from others. This is why an actor cannot conceal something from himself, and for this reason I think (iii) is correct behaviour, or certainly intentional behaviour.

I agree with Zed that (i) and (ii) are bugs, and that the inventory bug in (iv) should also be fixed. Since Zed has already drafted a fix, perhaps this could indeed become a PR when Zed's ready? (And would this enable us to close I7-2046 and I7-2036, in fact? Always good to reduce the pile.)

On (v), there are two reasons why rooms should not be able to conceal things. One is that if they did we would need to define concealment as a relation over (object, object), not (thing, thing), which would be slightly less efficient. This maybe isn't the most compelling reason in the world, but still. The other reason is that, again, concealment means concealing from the view of people outside. The example Peter gives of a door in a room doesn't make sense for that: if I'm in the room, I'm on the inside, and can see the door, which is only being concealed from people outside the room. (There are lots of ways to make hidden doors, but concealment isn't the right concept for it.)

(vi) is a subtle question. `concealed` as an adjective on its own can only be determined by knowing the point of view. If a packing crate conceals Houdini, and Houdini carries but does not conceal a key, and the player is also in the crate, should we say that that the key is `concealed`? The player can see it perfectly well; but in a sense it is concealed by the crate. It resolves ambiguities like this to say that `concealed`, the adjective, means `concealed by the holder`: in which case we say that the key is not. That reasoning seems right to me.

So I think (i), (ii), (iv) need addressing, and (iii), (v), (vi) should stay as they are.

**Comment by drpeterbatesuk** [comment]
I follow Graham's reasoning, given his different starting point as to the meaning of concealment.

That said, I feel it's perhaps an argument developed more from semantics than what would necessarily be helpful to an author.

One thing I would still take definite issue with is the issue of the concealment relation vs the definition of concealment. If we are to develop an argument from the idea that concealment depends on the point of view, then it becomes a relation with three parameters- the concealed item, the concealing item and the viewer. The current definition of concealed (_adjective_) takes no account at all of the viewer and strictly enforces a holder of/held item relationship between concealed and concealing things. That the concealment _relation_ allows an enclosed by/enclosed item relationship has no obvious functional purpose, since a thing something conceals but does not hold is not functionally concealed. Conversely, items enclosed by something that is concealed (adjective) _are_ functionally concealed even though in conditions they are not concealed (adjective) and nothing conceals them (relation). Neither out-of-the box have any idea of point-of view beyond an assumption that it's from outside the concealing thing.

The consequence of all this is that currently alignment of functional concealment with the concealment relation and the concealed adjective, and account if any of the point of view, must be achieved through particular phrasings of (or phrases in) 'Deciding the concealed possessions' rules, rather than being implemented in the World Model, e.g.:

>Rule for deciding the concealed possessions of  a \<description of things> called the hider when the holder of the particular possession is the hider and the hider does not enclose the player: decide yes.

Obviously, the 'when the holder of the particular possession is the hider' condition does the necessary here to align concealed and conceals, while 'the hider does not enclose the player' is a rudimentary stab at including point-of-view.

> It resolves ambiguities like this to say that concealed, the adjective, means concealed by the holder: 

Well, not entirely because Houdini is both concealed (adjective) and concealed by the crate, (unless the DTCP rules take care to consider the point of view of the player), yet he also is in full view of the player. Furthermore, since the key is enclosed by Houdini, who is concealed, functionally it too is concealed from the player although 'the key is concealed' is false.

One difficulty here is that if the same consideration that prevents rooms concealing anything is fundamental enough to the idea of concealment to be hard-wired into the World Model, the same should logically apply to things like the crate that enclose the point of view.  In the above example, the crate should not be allowed to conceal anything when the point of view is enclosed by it:

>The Stage is a room.
>The Escape Box is a  lit closed enterable openable container in the Stage. Houdini is a man in the escape box. Houdini carries a key. The player is in the escape box.

>**The Escape Box**
>In the Escape Box you can see Houdini.

>\>**x key**
You see nothing special about the key.

Add:
>Rule for deciding the concealed possessions of the escape box when the particular possession is Houdini: decide yes.

and:

>**The Escape Box**
>In the Escape Box you can see nothing.

>\>**x Houdini**
>You can't see any such thing.
>
>\>**x key**
>You can't see any such thing.

Change:

>Rule for deciding the concealed possessions of the escape box when the particular possession is Houdini and the holder of the player is not the escape box: decide yes.

and:

>**The Escape Box**
>In the Escape Box you can see Houdini.
>
>\>**open box**
>You open the Escape Box.
>
>\>**exit**
>You get out of the Escape Box.
>
>**Stage**
>You can see an Escape Box here.
>
>\>**search box**
>In the Escape Box is nothing.

I can't help thinking that the idea of concealment being something that can only occur when a point-of-view is from 'outside' the holder of the concealed item starts to struggle in particular when considering supporters. In terms of their ability to conceal, why should the floor (or walls) of a room be categorically  different from the surface of a table? Or, really, the inside of an opened matchbox? And if the player climbs onto a table of concealed secret documents, should they by default appear?

I guess that fundamentally I feel that this interpretation of concealment is not always a particularly helpful one to the author. It would perhaps be a middle ground to implement it in rules that can be abolished, rewritten or replaced rather than hard-wired into the World Model.  To be fair, (with the exception of direct possessions) that can already be done with the inventory issue, (which goes beyond a person not being able conceal something from themselves: if the player picks up a puzzle box concealing a diamond, the diamond is revealed).

**Comment by ganelson** [comment]
I agree that the bare word `concealed` is not perfect, but I think what we have (i.e., defining it as "concealed by the holder") is the best we can reasonably do without causing too much disruption. I think we should move on (except for fixing the issues (i), (ii), (iv), which are definitely valid).

**Comment by ganelson** [comment]
In the definition of enclosure, backdrops are now handled the same way
that two-sided doors are: just as both sides of the door enclose it, so all
rooms in which a backdrop is found enclose it. (Unless the backdrop is `absent`,
in which case they don't.)

**Comment by ganelson** [comment]
(That enclosure change caused no issues with the test suite, and means for the first time that holding implies enclosure.)

I did also investigate whether we could safely change room-containment so that a room doesn't room-contain itself. The answer is no, it's not safe to change that. The trouble is with conditions like `if R is somewhere dark`, which authors might intend either as "if R is a thing in a dark room" or "if R is itself a dark room", depending on whether R is a room or not. The only way to have it work both ways is for rooms to room-contain themselves; see the test case `WhereWords`.

**Comment by ganelson** [comment]
Something unfortunate which I should give notice of, though: Looking more carefully at how regions and backdrops are implemented, I think we have to take a firm stance that rooms cannot change region, and that regions cannot be moved to other regions, with `now`. I was always a little queasy about allowing this, but thought, well, why not. What I now realise is that it's much more problematic than it looks to do that, because of the `regional_found_in` property, where the compiler essentially represents the set of rooms in a region by compiling an indicator function for that set. If the regions can move around, that's going to fail, and result in subtle bugs where backdrops are still located where the previous region arrangement would suggest that they ought to be. So tomorrow I'll force regions and rooms to be immutable in the object tree.

**Comment by ganelson** [comment]
No, I take it back. On closer inspection, `regional_found_in` functions work through the regional-containment relation, and therefore do continue to work when regions or rooms are moved around. So I think we can indeed keep the ability to do this.

**Comment by zedlopez** [comment]
Just to be sure of desired behavior in the concealment (iv) case:

If a box is a container that conceals a note, and a player takes the box, then the default behavior will be that they will see nothing other than the usual "Taken." and the next time they do an inventory they'll see

```
a box
  a note
```

(all of which is the same as today) but now the note will be in scope, right?

**Comment by ganelson** [comment]
Okay, that seems wrong, doesn't it? I would have naively thought that the correct thing is for the note to be not only out of scope but also not listed in the inventory. Is there some reason why we don't want to do it that way?

**Comment by zedlopez** [comment]
I had thought the previous discussion indicated that what the player encloses shouldn't be concealed from the player. Is it merely that what the player directly holds isn't concealed from the player?

**Comment by ganelson** [comment]
Um, so if the player conceals the box, but the box does not conceal the note, then we should get the inventory:
```
a box
    a note
```
But if the box conceals the note, the inventory should surely be:
```
a box
```
I think the idea here is that concealment is about holders, not enclosers.

**Comment by drpeterbatesuk** [comment]
I think Graham's idea of this is what naive authors would instinctively expect, also what would be most useful to experienced authors, so I would support it on both those grounds.

Further, as Graham suggests, it fits with the idea that things (along with their contents) are functionally concealed only by their direct holder- so if actors can't conceal their direct possessions from themselves, likewise the contents of those possessions should be unconcealed unless the possessions themselves do their own concealing.

**Comment by drpeterbatesuk** [comment]
As things stand at present, it is already the case that concealment of things enclosed by an actor works in this way (which is, of course, why concealed things enclosed by the player, and their contents, are out of scope). It's just that at present by default inventory lists things concealed from the actor, and their contents. But I guess you knew that :-)

**Comment by zedlopez** [comment]
OK. That's the same behavior currently in Concealer and I agree that it's better (I was never advocating for the behavior I described above; I had misread Graham's comments as indicating it would be the desired behavior.)

**Comment by zedlopez** [comment]
I'm polishing the concealment code. I'm not sure what call to make for this case: The [Undescribed test case](https://github.com/ganelson/inform/blob/master/inform7/Tests/Test%20Cases/Undescribed.txt) has a glass jar with 3 undescribed scraps of paper inside. By the current behavior, they're not mentioned; if one searches or examines the jar, one is told it's empty. But if one drops another container alongside it, it'll have the '(empty)' tag in the room description and the jar won't, and, of course, the paper is in scope. On balance, I think it's probably better to have the '(empty)'. Any thoughts?

**Comment by ganelson** [comment]
Yes. It seems hard to argue with the point that a container with only undescribed contents should look the same as a container which is actually empty, which argues for printing the '(empty)', even though in some sense this is a lie. It's weird to be effectively transmitting side-channel information to the player who is sufficiently au fait with Inform's internals, i.e., that because we don't print '(empty)' now we are in effect signalling that the container must contain something undescribed.

It's interesting that we used the term "undescribed" and not, say, "invisible". At some point it was conceptually a little less clear-cut than simply not being seen. This no doubt goes back to the "concealed" attribute in Inform 1.

**Comment by zedlopez** [comment]
Revenge of the World Model!

I found another way a person can end up enclosing a person.

```
lab is a room.
parlor is north of lab.

the can't push people rule does nothing.
the player is pushable between rooms.

Gallant is a person in the lab.
Persuasion rule for asking Gallant to try doing something: persuasion succeeds.
```
->
```
gallant, push me north
Gallant arrives from the south, pushing yourself in front, taking you along.

parlor (in Gallant)

>
```
OK, but it took deactivating a standard rule, so maybe it doesn't count. But...
```
lab is a room.
parlor is north of lab.

sled is an enterable pushable between rooms supporter in lab.
the player is on sled.

Gallant is a person in the lab.
Persuasion rule for asking Gallant to try doing something: persuasion succeeds.
```
->
```
>gallant, push sled north
Gallant goes north, pushing sled in front, taking you along.

parlor (on sled) (in Gallant)
parlor (on sled)
You can see Gallant here.

>
```

The actor becomes the holder of the thing being pushed, and thus encloses the player during the implicit look that's part of the going action below...
```
[ ConvertToGoingWithPush i oldrm newrm infl;
        i=noun;
        if (IndirectlyContains(noun, actor) == false) { move i to actor; infl = true; }
        move_pushing = i;
        oldrm = LocationOf(noun);
        BeginAction(##Go, second);
        newrm = LocationOf(actor);
        move_pushing = nothing; move i to newrm;
        if (newrm ~= oldrm) {
                if (IndirectlyContains(i, player)) TryAction(0, player, ##Look, 0, 0);
                RulebookSucceeds();
        } else RulebookFails();
        rtrue;
];
````
this seems like a reasonable and parsimonious patch for this, but I'm testing further. (Found this while working on fixing [I7-1995](https://inform7.atlassian.net/browse/I7-1995).)
```
before looking when the player is enclosed by something enclosed by a person who is not enterable: instead do nothing.
```

**Comment by ganelson** [comment]
I'm shocked, shocked I say. That would probably win the prize for the most eldritch rule in the Standard Rules, but it seems a reasonable solution to me. Would we give it a name? (I cannot imagine what.)

**Comment by erkyrath** [comment]
> This no doubt goes back to the "concealed" attribute in Inform 1.

It certainly does, and "undescribed" remains untidy to this day. (SCORE__UNCONCEALED in Adjudicate(), etc.)

**Comment by zedlopez** [comment]
> now X does not hold Y is never permitted.

This is problematic for the abstract object case: there's no non-I6 way to undo the holding relation.

**Comment by zedlopez** [comment]
This may have already been argued as much as that most controversial "describe what's on scenery supporters in room descriptions rule", so my apologies if revisiting it is unwelcome. If you search a container or supporter (the container or supporter in question can be scenery or not), held scenery things are equivalent to undescribed or concealed things: unless there's something directly in or on the thing that's none of scenery, undescribed, concealed, then upon searching it, the default is that the player is told "[The container] is empty" or "There is nothing on [the supporter]". The same "[The container] is empty" message is given when examining containers that contain scenery but either contain nothing else or only contain other things that are concealed or undescribed.

This seems off to me for the scenery case. Scenery is *de-emphasized* but it seems weird to explicitly *deny* its existence. I think in the searching case that the default should be that the player is told about scenery things in or on the thing being searched (and they hold that are neither concealed nor undescribed, and so on) or in containers being examined.

I'm not suggesting changing any of looking's behavior, or anything else about examining's or searching's behavior in this regard.

**Comment by erkyrath** [comment]
> Scenery is de-emphasized but it seems weird to explicitly deny its existence.

Hm. Scenery is (as a rule) not hidden from the player. The point of `scenery` is that it's already accounted in the room description, so it should be hidden *from the LOOK action*.

That doesn't mean we should hide a scenery object from the SEARCH action. However, both LOOK and SEARCH are going to do a bit of "list the contents of the container" business. It would be both fussy and surprising if that contents list included or excluded scenery objects depending on the current action.

If I recall correctly, that "scenery is equivalent to concealed" logic was added because of a bug report reading like:

> The trophy case contains.

(All contents were scenery, therefore all skipped.)

**Comment by erkyrath** [comment]
To be sure, I've long since decided to only use `scenery` for top-level objects, never for objects in a container or supporter. For exactly these sorts of reasons!

**Comment by zedlopez** [comment]
I just noticed another weirdness of the existing concealment behavior that my fixes in 10.2 didn't change:

an item can be all of concealed, lit, and the only light source in the room, and it'll remain concealed but the player won't be in darkness.

My thinking is that the room should remain in darkness. Certainly one could say that it's more "realistic" that a lit concealed item in a dark room would thus reveal itself, but that would mean trumping the author's concealment rules, which I wouldn't want to do.

But I think if concealment-trumps-litness is chosen and implemented the docs should call attention to being the case that if an author creates a situation where a concealed lit item might be the only light source in a dark room, they're on their own regarding what to do about it.

**Comment by zedlopez** [comment]
@ganelson My draft outlining further suggestions for World Model enforcement is now available in a branch on my fork of this repo. (Fear not, everyone; I *won't* be making a pull request.)

https://github.com/zedlopez/inform-evolution/blob/world_model/proposals/0015-world-model-enforcement.md#implicit-taking

**Comment by CelticMinstrel** [comment]
The description of changes to the support relation sounds like it would break the Rideable Vehicles extension, since it redefines the relation in terms of a supporter, yet neither rideable vehicle nor rideable animal are kinds of supporter. Was this tested?

**Comment by zedlopez** [comment]
Rideable vehicles *are* kinds of supporter.
Rideable animals aren't, but they have the I6 supporter flag.

There *are* problems with rideable animals. These work:

- try mounting the horse;
- move the player to the horse;
- if the holder of the player is the horse...
- if the horse holds the player...
- if the horse encloses the player...

This doesn't:

`now the player is on the horse`

> *** Run-time problem UsedSpatialRelationOnNonThing: INTERNAL/Inter/WorldModelKit/RTPs
>*** Holding, wearing and carrying are for things only.
>*** Attempted to move yourself to foo.

Testing `if the horse supports the player ` or `if the player is on the horse` compiles and runs without complaint, but is always false.

**Comment by zedlopez** [comment]
You can still `move x to y` in the general case, including moving the player to any thing that isn't off-stage. This is a feature, because it lets us use the object tree in subkinds of object, but maybe MoveObject should block moving anything `ofclass K2_thing` to anything but a room, or something with the supporter or container flags, which would make rideable animals (which have the supporter flag) remain ok (and there would still be an exception to block moving the player to nowhere or to anything off-stage, of course).

Additional work would be necessary to make the support relation work.

**Comment by zedlopez** [comment]
Hmm. If rideable animal were a kind of supporter that was usually animate, male, not neuter, the only things I can think of that you'd lose would be: 
- a rideable animal couldn't relate to things by carrying, wearing, possession; tests would need to be in terms of "supports / is on"
- wouldn't be a person, so wouldn't trigger many check rules involving other people
- couldn't assign a rideable animal to the player variable
The loss of wearing might be a pain point, but the others don't seem so bad.

**Comment by zedlopez** [comment]
actually, I was confused. There isn't a problem. *It's a bug in versions before current development* that one **can** say `now the player is on the horse` because the support relation never applied. `if the horse supports the player` *should* always be false. In both 10.1 and current development, `now the horse carries the player` and `if the horse carries the player` work.

**Comment by CelticMinstrel** [comment]
Is it at all possible to give "on" a different meaning when used with rideable animals? Right now it looks like the compiler rejects using the same verb for two different relations, but it would definitely be convenient if "on the horse" were translated to the carrying relation, and with the support relation defined only on supporters, it's not as if that would actually be ambiguous (it can be disambiguated in the same way that the containment and regional-containment relation already is).

I suppose it's not that big of a deal if that can't be done or is just too complicated. That said, the Rideable Animals documentation should maybe mention explicitly that rideable animals carry you rather than support you.

**Comment by zedlopez** [comment]
> Is it at all possible to give "on" a different meaning when used with rideable animals?

No. A perhaps helpful sort-of-mitigation might be:

```
ridership relates a thing (called rider) to a thing (called ridden) when ridden is vehicular and ridden holds rider.

The verb to ride means the ridership relation.
The verb to ride in means the ridership relation.
The verb to ride on means the ridership relation.
```

with that `player rides the horse`, `horse is ridden by the player`, `player rides on the unicycle`, `player rides in the popemobile`  all work as expected. And then RV could encourage use of ride/ridden by rather than carries/carried by or supports/is on.

CelticMinstrel:
> And then RV could encourage use of ride/ridden by rather than carries/carried by or supports/is on.

That sounds great, except that the relation is not assertable, right? So RV would still have to make it clear that rideable vehicles support their passenger(s) while rideable animals carry them.

zedlopez:
Yes, it could only be used in conditionals, not in assertions or with `now`.

PR #17, opened 2023-05-02
IE-0017: Apps and extensions

zedlopez:
Proposal: [IE-0017](https://github.com/ganelson/inform-evolution/blob/main/proposals/0017-apps-and-extensions.md)
Authors: AUTHOR
Language feature name: --
Status: Draft
Related proposals: [IE-0001](https://github.com/ganelson/inform-evolution/blob/main/proposals/0001-extensions-with-resources.md)
Implementation: In progress

### DIFF
--- a/proposals/0017-apps-and-extensions.md
+++ b/proposals/0017-apps-and-extensions.md
@@ -113,12 +113,12 @@ The apps must be able to install from the following sources:
 
 We may eventually want to add two other capabilities, but are not sure:
 * A drag of an extension file (or directory) onto the project window, or perhaps just onto its toolbar or one of the tabs other than "Source". (Drags into source generally paste in the contents of text files.)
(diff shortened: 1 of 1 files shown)


ptomato:
I'm working on implementing the GUI part of this. It'd be convenient if inbuild would generate the confirmation page such that clicking the confirm button would call `project().confirmAction(...)` with one argument (instead of zero as it currently does). The argument would be the exact path that was given with the `-install` argument on inbuild's command line.

That way, the confirmation flow could be stateless.

Currently, the app has to remember the path of the extension it most recently tried to install with inbuild. This is not a huge problem because I don't _think_ it's currently possible to get the app into a state where the Report tab is showing a confirmation page that refers to a different extension than the app's internal state containing the last remembered extension, but statelessness would preclude any bugs of this nature altogether.

PR #27, opened 2023-07-19
IE-0027: Sorting with custom comparison phrases

curiousdannii:
# (IE-0027) Sorting with custom comparison phrases

* Proposal: [IE-0027](https://github.com/ganelson/inform-evolution/blob/main/proposals/0027-sorting-with-phrases.md)
* Author: Dannii Willis
* Language feature name: None
* Status: In progress
* Related proposals: None

## Summary

Add phrases that allow lists and tables to be sorted with custom comparison
phrases.


CelticMinstrel:
> The list phrase functions just as other common languages do (for example, see [C#](https://learn.microsoft.com/en-us/dotnet/api/system.collections.icomparer.compare?view=net-7.0#system-collections-icomparer-compare(system-object-system-object)), [Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)). The phrase takes two list elements and returns a number indicating the relative order of the elements: negative if the left value should come before the right, zero if they are equal, and positive if the left value should come after the right.

I think it might be better to take [the new C++ comparison system](https://en.cppreference.com/w/cpp/utility/compare/partial_ordering) as a model and have the comparator functions return an enumerated type with either three or four possible values (depending on whether you want sort to work on a partial order). In particular, this cuts off the common but incorrect practice of returning the difference of two numbers as the comparison result. (It's incorrect because of overflow – so it _can_ be correct, but only with an unlimited precision bignum type.)

> Since there is no kind representing an entire table row, to sort a table we pass the table name and two row numbers. The comparison phrase can then access any columns it wishes in order to determine the relative order of the two rows. Any completely blank rows will be put at the bottom, just as with the existing sorting phrases.

SQL offers an option to put blank rows (or rather, nulls) at the top instead, is there any benefit to that? Maybe there isn't, but I just thought it was worth asking. If desired it could just be a phrase option ("blanks first or blanks last). (This question actually applies to the pre-existing sorts as well.)

Side note while I'm here: the documentation doesn't state whether sorts of lists are stable or not. This seems like a pretty important thing to know. It does mention that sorting a _table_ is stable, but I think you should be explicit about it in all cases.

PR #29, opened 2023-07-20
IE-0029: Basic IO

curiousdannii:
# (IE-0029) Basic IO

* Proposal: [IE-0029](https://github.com/ganelson/inform-evolution/blob/main/proposals/0029-basic-io.md)
* Author: Dannii Willis
* Language feature name: None
* Status: Accepted
* Related proposals: #17

## Summary

Extend the basic IO functionality of Basic Inform, including incorporating Basic Screen Effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,5 +60,4 @@ _(IE-0011) New data structures_ (still drafting: see pull request)              
 [(IE-0026) Units and number bases](proposals/0026-units-and-number-bases.md)                             | 13 July 2023    | In progress                | [PR#26](https://github.com/ganelson/inform-evolution/pull/26)
 [(IE-0027) Sorting with custom comparison phrases](proposals/0027-sorting-with-phrases.md)               | 20 July 2023    | In progress                | [PR#27](https://github.com/ganelson/inform-evolution/pull/27)
 [(IE-0028) Extension documentation revisited](proposals/0028-extension-documentation-revisited.md)       | 19 July 2023    | In progress                | [PR#28](https://github.com/ganelson/inform-evolution/pull/28) 
-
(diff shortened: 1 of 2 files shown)


curiousdannii:
One question is how to handle number input. Inform doesn't (currently) have an option kind, so we can't return a failure. (Unless we returned it in a separate value? Very messy.) Therefore we probably want to make the phrase ask the player a second time if their input wasn't a valid answer. That means reprinting the prompt as well. So maybe something like

> ask the player for a number with prompt (P - text)

ganelson:
Isn't it better to have a phrase which inputs an arbitrary text, which can then be composed with a phrase to convert text to a number?

But there ways using inline definitions to make conditional assignments of the "if let" sort, if you reaaally want to.

curiousdannii:
Ah, yeah that would probably be a better option. The author could use regexes to test the string is only a number (optionally wrapped in spaces), and if that check passes then parse it into a number. If it doesn't match then the author can print the prompt and ask again themselves. So then we'd only need one phrase for text input, but then a separate phrase to parse a text into a number. And maybe also to parse into a real number?

If let... I do really like them, but when I've tried it before it's been really hacky. Something to consider for the future.

curiousdannii:
I've begun working on this, and have more thoughts.

## Status line
BSE's status line phrases are basically completely disconnected from how Inform usually displays the status line, and in particular from the constructing the status line activity.

The activity should be moved from the standard rules into basic inform. The basic IO phrase(s) would then be one particular way of using the activity.

## Picking a number
DialogKit defines `DirectorPickANumber`. There's a lot of overlap between this and the suggested number parsing function...

PR #30, opened 2023-08-09
IE-0030: Extension examples and testing

zedlopez:
Proposal: [IE-0030](https://github.com/ganelson/inform-evolution/blob/main/proposals/0030-extension-examples-and-testing.md)
    Authors: Graham Nelson
    Language feature name: --
    Status: Draft
    Related proposals: [IE-0028](https://github.com/ganelson/inform-evolution/blob/main/proposals/0028-extension-documentation-revisited.md), [IE-0001](https://github.com/ganelson/inform-evolution/blob/main/proposals/0001-extensions-with-resources.md)
    Implementation: None as yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Proposal                                                                        
 Proposal                                                                                                 | Began           | Comments
 -------------------------------------------------------------------------------------------------------- | --------------- | --------
 [(IE-0031) BANISH debugging command](proposals/0031-banish.md)                                           | 9 August 2023   | [PR#31](https://github.com/ganelson/inform-evolution/pull/31) 
-[(IE-0029) Extension examples and testing](proposals/0029-extension-examples-and-testing.md)             | 7 August 2023   | [PR#28](https://github.com/ganelson/inform-evolution/pull/29) 
+[(IE-0029) Extension examples and testing](proposals/0029-extension-examples-and-testing.md)             | 7 August 2023   | [PR#29](https://github.com/ganelson/inform-evolution/pull/29) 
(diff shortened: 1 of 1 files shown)


zedlopez:
This would be a really nice, I'm guessing very easy tweak...

For the I7 test source files, instead of:
```
Test: testname
For: compilation-target
```

make it...

```
[ Test: testname ]
[ For: compilation-target ]
```

and then they can be compiled without any need to trim the header.

zedlopez:
When one specifies multiple -show flags at once, do they apply to the same compilation sequence, or does it start over fresh with each one? The former would be a nice time-saver if it's not already the case.

It'd also be nice if the Inform6 compilation didn't use -x.

intest is rejecting -internal.

../intest/Tangled/intest -internal inform7/Internal ~/mine/extensions/AW\ Freyr/Hybrid\ Choices-v7_1.i7xd/ -test all
intest: fatal error: no such test case as /home/zed/mine/extensions/AW Freyr/Hybrid Choices-v7_1.i7xd/

I think it's important for extension testing that intest be able to take a -nest parameter, what with extensions commonly having dependencies on other extensions.

ganelson:
I agree about removing -x, and have done that.

I don't think intest should literally support -internal because it isn't tied to Inform in quite so close a way as that. The main.delia testing script does support changing the internal nest with the $$internal variable, which can be set on the command line. Still, that's all rather clunky.

One of my intentions for the next release cycle of Inform is to write a "driver", that is, a better and more consistent command-line interface for the Inform toolchain, using subcommands like "inform build X" and "inform test X" to drive the other tools - this seems now to be considered good CLI practice, and would clarify the currently rather difficult to understand overlapping sets of options on the tools.

ganelson:
On the test file headers, I think that would be an excellent suggestion if we only had test files, but we also have example files, and I want the two to be consistent with each other. That is, I think a test file should model itself more on an example file than on an I7 source file as such.

PR #34, opened 2023-10-29
IE-0034: A new Glulx random algorithm

curiousdannii:
* Proposal: [IE-0034](https://github.com/ganelson/inform-evolution/blob/main/proposals/0034-glulx-random.md)
* Authors: Dannii Willis
* Status: Accepted
* Implementation: Not yet implemented

## Summary

Existing Glulx interpreters can have flawed random implementations, so move the
implementation of Inform's `random` function inside Architecture32Kit so that
we can ensure it is reliable.

### DIFF
--- a/proposals/0034-glulx-random.md
+++ b/proposals/0034-glulx-random.md
@@ -1,0 +1,131 @@
+# (IE-0034) A new Glulx random algorithm
+
+* Proposal: [IE-0034](0034-glulx-random.md)
+* Discussion PR link: [#34](https://github.com/ganelson/inform-evolution/pull/34)
+* Authors: Dannii Willis
+* Status: Accepted
+* Implementation: Not yet implemented
+
+## Summary
+
+Existing Glulx interpreters can have flawed random implementations, so move the
+implementation of Inform's `random`` function inside Architecture32Kit so that
+we can ensure it is reliable.
+
+## Motivation
+
+In September 2023 it was revealed in [a forum discussion](https://intfiction.org/t/lack-of-randomness-sometimes-when-compiling-for-glulx/64533)
+that some Glulx interpters have very flawed random number generators. In particular,
+when requesting a random number between 1 and N, if N is a power of 2, then there
+will only be N distinct sequences. Even worse, the initial numbers of those sequences
+are often the same, the example code showing that when requesting a random number
+between 1 and 4, the first number will always be a 3. This could have dire
+consequences: if a murder mystery game picked randomly between 4 suspects, in these
+interpreters it would always pick the third suspect.
+
+The affected interpreters include Glulxe in Windows, and Git in all OSes. These
+interpreters have since been patched to use much more reliable RNGs, but it can
+take a long time for downstream interpreters to include these updates; many
+interpreters are only updated infrequently, and some are no longer maintained,
+including all Android Glulx interpreters.
+
+Inform must provide high quality randomness out-of-the-box without authors needing
+to dig into interpreter quirks. But as we can't rely on the interpreters being
+patched in a timely manner, we need to instead move the implementation of Inform's
+RNG inside Inform kit code (Architecture32Kit) to insulate authors from the flaws
+of the interpreters. Many other programming languages also implement their RNGs
+inside their library code rather than relying on the OS-provided RNG, so Inform
+will be in good company doing this.
+
+## Components affected
+
+- [ ] No change to the natural-language syntax.
+- [ ] No change to inbuild.
+- [x] Minor changes to inform7.
+- [ ] No change to inter.
+- [ ] No change to the Inter specification.
+- [x] Minor changes to runtime kits.
+- [ ] No changes to the Standard Rules and Basic Inform.
+- [ ] No change to documentation.
+- [ ] No change to the GUI apps.
+
+## Impact on existing projects
+
+Authors will receive better quality random numbers without any change to their
+code. However if anyone is unintentionally relying on the quirky results of
+calling `random` with a negative value in Glulx, they will need to change what
+they are doing.
+
+## Details
+
+The Inform 6 function `random` is a "system" function. This means that by default
+it is not a true function, but instead just a shortcut for the VM-specific random
+opcode. System functions can however be replaced by the author, so that is what
+we will do to fix this problem, replacing it with a function in Architecture32Kit.
+
+Considerations:
+
+1. Ideally we want a RNG that produces high quality random numbers, has decently
+    high performance, and does not use excessive amounts of memory. The [Xorshift](https://en.wikipedia.org/wiki/Xorshift)
+    family of RNGs fit these criteria. Glulxe and Git switched to the
+    [xoshiro128** algorithm](https://prng.di.unimi.it/), so we might as well
+    use it too.
+
+2. This RNG can be seeded with any non-zero value, and it is common to seed it
+    with a timestamp. The Inform test suite however requires predictable randomness,
+    which it currently accomplishes through Glulxe's `--rngseed` argument. So
+    rather than seeding Inform's new RNG with a timestamp, we should instead seed
+    it with `@random`. Even though the interpreter's RNG may be flawed, one call
+    with a high enough range is probably much safer than calling it with a range
+    of only 4 or 8. It has not yet been determined whether a full 32-bit `@random`
+    call is safe in affected interpeters, or instead whether it should make a
+    anged call using a high prime number, or if any priming is necessary.
+
+3. We need to replace the Inform 6 `random` function rather than introducing a
+    new random function so that all randomness, even in extensions, uses the new
+    algorithm.
+    
+    But the I6 `random` function does more than just returning a random number.
+    As [documented in the DM4](https://www.inform-fiction.org/manual/html/s1.html#p45)
+    `random` can also be called with multiple arguments, for it to randomly choose
+    between. This however [does not work in Inform 7](https://intfiction.org/t/lack-of-randomness-sometimes-when-compiling-for-glulx/64533/101).
+    Inline invocations like this
+    ```
+    To decide which number is todays number: (- random(1,2,3,4,5,6,7,8,9) -).
+    ```
+    produce the error
+    > inform7: this inv of !random should have 1 argument(s), but has 9
+    
+    But calls to `random` in whole-function inclusions like this
+    ```
+    Include (-
+    [ TodaysNumber;
+        return random(1,2,3,4,5,6,7,8,9);
+    ];
+    -).
+    ```
+    compile, however they are silently modified during the Inter transformation
+    to only include one argument in the final build.
+
+    Inform 7 should consistently reject the multiple-argument form of `random`
+    at all times.
+
+4. There is a further compilication: the Glulx I6 implementation of `random`
+    does not match the behaviour described in the DM4. Instead of negative or
+    zero values seeding the interpreter's RNG, it is passed directly to [the Glulx
+    `@random` opcode](https://www.eblong.com/zarf/glulx/Glulx-Spec.html#opcodes_rand),
+    before adding 1. This results in the following inconsistencies:
+
+    | Argument | DM4/Z-Machine | Glulx |
+    |----------|---------------|-------|
+    | Positive N | Return a random number in the range 1 to N | Return a random number in the range 1 to N |
+    | Zero | Switch the interpreter RNG to random mode | Return a full 32-bit random number |
+    | Negative N | Seed the interpreter RNG in predictable mode | Return a random number in the range (N + 2) to 1
+
+    Nobody would intentionally call `random(-10)` expecting it to return a number
+    between -8 and 1, so we don't need to be concerned with preserving the I6
+    behaviour of `random` in Glulx; instead we should make it conform to the DM4.
+
+5. People may find the behaviour of the Glulx `@random` opcode more useful than
+    the I6 `random` function, so we should also provide a function that mimics
+    the `@random` opcode, but using the new RNG.
--- a/proposals/0034-glulx-random.md
+++ b/proposals/0034-glulx-random.md
@@ -81,7 +81,12 @@ Considerations:
     call is safe in affected interpreters, or instead whether it should make a
     ranged call using a high prime number, or if any priming is necessary.
 
-3. We need to replace the Inform 6 `random` function rather than introducing a
+3. Unless the new RNG is in predictable mode, it will need to be reseeded from
+    `@random` after restoring or undoing. This means an extra variable will be
+    needed to record whether the RNG is in 'random' or 'predictable' mode, as
+    the RNG's own state would otherwise be indistinguishable.
+
+4. We need to replace the Inform 6 `random` function rather than introducing a
     new random function so that all randomness, even in extensions, uses the new
     algorithm.
     
@@ -110,7 +115,7 @@ Considerations:
     Inform 7 should consistently reject the multiple-argument form of `random`
     at all times.
 
-4. There is a further complication: the Glulx I6 implementation of `random`
+5. There is a further complication: the Glulx I6 implementation of `random`
     does not match the behaviour described in the DM4. Instead of negative or
     zero values seeding the interpreter's RNG, it is passed directly to [the Glulx
     `@random` opcode](https://www.eblong.com/zarf/glulx/Glulx-Spec.html#opcodes_rand),
@@ -126,9 +131,9 @@ Considerations:
     between -8 and 1, so we don't need to be concerned with preserving the I6
     behaviour of `random` in Glulx; instead we should make it conform to the DM4.
 
-5. People may find the behaviour of the Glulx `@random` opcode more useful than
+6. People may find the behaviour of the Glulx `@random` opcode more useful than
     the I6 `random` function, so we should also provide a function that mimics
     the `@random` opcode, but using the new RNG.
 
-6. The Inform-to-C mode should also be checked and updated so that its `random`
+7. The Inform-to-C mode should also be checked and updated so that its `random`
     and `@random` are reliable.
--- a/proposals/0034-glulx-random.md
+++ b/proposals/0034-glulx-random.md
@@ -9,7 +9,7 @@
 ## Summary
 
 Existing Glulx interpreters can have flawed random implementations, so move the
-implementation of Inform's `random`` function inside Architecture32Kit so that
+implementation of Inform's `random` function inside Architecture32Kit so that
 we can ensure it is reliable.
 
 ## Motivation
--- a/proposals/0034-glulx-random.md
+++ b/proposals/0034-glulx-random.md
@@ -129,3 +129,6 @@ Considerations:
 5. People may find the behaviour of the Glulx `@random` opcode more useful than
     the I6 `random` function, so we should also provide a function that mimics
     the `@random` opcode, but using the new RNG.
+
+6. The Inform-to-C mode should also be checked and updated so that its `random`
+    and `@random` are reliable.
--- a/proposals/0034-glulx-random.md
+++ b/proposals/0034-glulx-random.md
@@ -15,13 +15,13 @@ we can ensure it is reliable.
 ## Motivation
 
 In September 2023 it was revealed in [a forum discussion](https://intfiction.org/t/lack-of-randomness-sometimes-when-compiling-for-glulx/64533)
-that some Glulx interpters have very flawed random number generators. In particular,
-when requesting a random number between 1 and N, if N is a power of 2, then there
-will only be N distinct sequences. Even worse, the initial numbers of those sequences
-are often the same, the example code showing that when requesting a random number
-between 1 and 4, the first number will always be a 3. This could have dire
-consequences: if a murder mystery game picked randomly between 4 suspects, in these
-interpreters it would always pick the third suspect.
+that some Glulx interpreters have very flawed random number generators. In
+particular, when requesting a random number between 1 and N, if N is a power of
+2, then there will only be N distinct sequences. Even worse, the initial numbers
+of those sequences are often the same, the example code showing that when
+requesting a random number between 1 and 4, the first number will always be a 3.
+This could have dire consequences: if a murder mystery game picked randomly
+between 4 suspects, in these interpreters it would always pick the third suspect.
 
 The affected interpreters include Glulxe in Windows, and Git in all OSes. These
 interpreters have since been patched to use much more reliable RNGs, but it can
@@ -65,9 +65,9 @@ we will do to fix this problem, replacing it with a function in Architecture32Ki
 
 Considerations:
 
-1. Ideally we want a RNG that produces high quality random numbers, has decently
+1. Ideally we want an RNG that produces high quality random numbers, has decently
     high performance, and does not use excessive amounts of memory. The [Xorshift](https://en.wikipedia.org/wiki/Xorshift)
-    family of RNGs fit these criteria. Glulxe and Git switched to the
+    family of RNGs fits these criteria. Glulxe and Git switched to the
     [xoshiro128** algorithm](https://prng.di.unimi.it/), so we might as well
     use it too.
 
@@ -78,8 +78,8 @@ Considerations:
     it with `@random`. Even though the interpreter's RNG may be flawed, one call
     with a high enough range is probably much safer than calling it with a range
     of only 4 or 8. It has not yet been determined whether a full 32-bit `@random`
-    call is safe in affected interpeters, or instead whether it should make a
-    anged call using a high prime number, or if any priming is necessary.
+    call is safe in affected interpreters, or instead whether it should make a
+    ranged call using a high prime number, or if any priming is necessary.
 
 3. We need to replace the Inform 6 `random` function rather than introducing a
     new random function so that all randomness, even in extensions, uses the new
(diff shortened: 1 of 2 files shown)


proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+The affected interpreters include Glulxe in Windows, and Git in all OSes. These
+interpreters have since been patched to use much more reliable RNGs, but it can
+take a long time for downstream interpreters to include these updates; many
+interpreters are only updated infrequently, and some are no longer maintained,
+including all Android Glulx interpreters.
+
+Inform must provide high quality randomness out-of-the-box without authors needing
+to dig into interpreter quirks. But as we can't rely on the interpreters being
+patched in a timely manner, we need to instead move the implementation of Inform's
+RNG inside Inform kit code (Architecture32Kit) to insulate authors from the flaws
```
hanna-kruppe, line 35: What about Architecture16Kit (i.e., Z-Code)? I don't know if the RNG quality of those terps is a concern (I recall it was briefly discussed in the forum thread, but I don't think there was any definite conclusion), but if it is, it's probably better to address it at the same time.

proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+opcode. System functions can however be replaced by the author, so that is what
+we will do to fix this problem, replacing it with a function in Architecture32Kit.
+
+Considerations:
+
+1. Ideally we want an RNG that produces high quality random numbers, has decently
+    high performance, and does not use excessive amounts of memory. The [Xorshift](https://en.wikipedia.org/wiki/Xorshift)
+    family of RNGs fits these criteria. Glulxe and Git switched to the
+    [xoshiro128** algorithm](https://prng.di.unimi.it/), so we might as well
+    use it too.
```
hanna-kruppe, line 72: Arguably, the bar for confidence in and future proofing of RNG quality should be even higher if the algorithm is to be baked into storyfiles. I've not even fully convinced myself of this, but I can see some good arguments for it. Updating old games (and getting people to play the new build instead of the old one) is even harder than updating terps and getting them into everyone's hands, so it's more important to get this right in the first try. And while baking an RNG into the storyfile can address old terps' RNG issues, it also means an RNG baked into storyfiles that later turns out to be flawed can't be fixed by just using a new terp with a better, or at least different, RNG implementation.

The big hammer for addressing such concerns would be a cryptographically secure PRNG algorithm, whose statistical quality is judged by a far higher standard (and tested much more thoroughly) than that of other PRNGs. Security is not relevant in this context, but it implies "indistinguishable from true randomness" (if the seed/internal state is unknown), which is stronger than necessary for IF but definitely not a bad thing. These RNG designs are a bit unwieldy to integrate because they are often based on block ciphers and thus generate small batches of random numbers instead of one number at a time, but I don't think something like ChaCha12 would be too slow or too bulky to be acceptable.

proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+
+2. This RNG can be seeded with any non-zero value, and it is common to seed it
```
hanna-kruppe, line 74: Seeding requested by the game via I7 phrases or I6 `random(something negative)` is restricted to a single number, but on all other occasions we can get a larger seed from multiple `@random` instructions to initialize the RNG's full internal state more or less directly. This may be desirable because it would make it extremely unlikely for such an "unpredictable" RNG state to coincide with the state from any particular 32-bit seed.

(However, for some non-crypto RNGs, a low quality seed can do much more damage in this scenario. For example, an all-zeros state is catastrophic for Xorshift-style algorithms, and states with low numbers of 1 bits are also undesirable.)

proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+    with a timestamp. The Inform test suite however requires predictable randomness,
+    which it currently accomplishes through Glulxe's `--rngseed` argument. So
+    rather than seeding Inform's new RNG with a timestamp, we should instead seed
+    it with `@random`. Even though the interpreter's RNG may be flawed, one call
+    with a high enough range is probably much safer than calling it with a range
+    of only 4 or 8. It has not yet been determined whether a full 32-bit `@random`
+    call is safe in affected interpreters, or instead whether it should make a
+    ranged call using a high prime number, or if any priming is necessary.
```
hanna-kruppe, line 82: Asking for a random number up to some large (close to 2^32, but not a power of two itself) N runs into another RNG quality issue in terps that implement "random number between 0 and positive N" as `random_uint32() % N`. This method introduces some bias in the distribution of the results, [because some remainders are more common than others (unless N is a power of two)](https://stackoverflow.com/a/10984975). For N much smaller than 2^32, this bias is usually negligible, but it gets progressively worse as N grows (never quite as pathological as the issues that launched this proposal, but still quite notable). Unfortunately, both old and updated versions of [Git](https://github.com/DavidKinder/Git/blob/71f94ec949231a87be4e795513c800e22e765886/terp.c#L1319-L1328) and [Glulxe](https://github.com/erkyrath/glulxe/blob/84f21bd7afb77532c211788de8e04068c6c286b6/exec.c#L659-L668) do this, and the updated [Quixe also does it](https://github.com/erkyrath/quixe/blob/939e03bffc763497f00de854ca71879f09ad3fb5/src/quixe/quixe.js#L4336-L4343) when it's using its xoshiro128\*\* RNG.

(Avoiding this problem is another way in which an Inform-internal RNG implementation could improve on what terps in the wild do, but for now it complicates the seeding story.)

proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+- [ ] No change to inbuild.
+- [x] Minor changes to inform7.
+- [ ] No change to inter.
+- [ ] No change to the Inter specification.
+- [x] Minor changes to runtime kits.
+- [ ] No changes to the Standard Rules and Basic Inform.
+- [ ] No change to documentation.
+- [ ] No change to the GUI apps.
+
+## Impact on existing projects
```
hanna-kruppe, line 52: I don't see any discussion of *when* to seed the RNG. Just doing it at game startup (which is where my mind went first) would mean that RNG state is part of quetzal saves and undo states: repeatedly reloading / undoing and repeating a turn would always reproduce the same random outcomes. This is probably undesirable for many people, others may welcome it, but in any case it would be be a marked change of behavior compared to the Glulx-native RNG, which is typically not saved and restored. I think this can be avoided by re-seeding from `@random` after every restore / restoreundo / restart / etc. (and autosave/autorestore is in the hands of the terp as usual), but I'm not 100% sure so I'm raising this more as a question than a comment.

proposals/0034-glulx-random.md
```diff
@@ -0,0 +1,134 @@
...
+    |----------|---------------|-------|
+    | Positive N | Return a random number in the range 1 to N | Return a random number in the range 1 to N |
+    | Zero | Switch the interpreter RNG to random mode | Return a full 32-bit random number |
+    | Negative N | Seed the interpreter RNG in predictable mode | Return a random number in the range (N + 2) to 1
+
+    Nobody would intentionally call `random(-10)` expecting it to return a number
+    between -8 and 1, so we don't need to be concerned with preserving the I6
+    behaviour of `random` in Glulx; instead we should make it conform to the DM4.
+
+5. People may find the behaviour of the Glulx `@random` opcode more useful than
```
hanna-kruppe: I'm not sure if it has to match `@random` exactly (its semantics are a bit awkward due to being essentially overloaded), but I'd definitely suggest an equivalent of `@random 0`, getting a full 32 bits of randomness. That's a very useful primitive for all sorts of things the standard rules don't and shouldn't provide, any underlying RNG can support it for free, and having to build it from the behavior of I6 `random()` or Glulx `@random nonZero` is possible but an unnecessary contortion.

proposals/0034-glulx-random.md
curiousdannii: Good point. We could make it reseed the RNG during GGRecoverObjects.

proposals/0034-glulx-random.md
curiousdannii: Matching the full `@random` opcode won't be difficult.

proposals/0034-glulx-random.md
curiousdannii: Hmm good point. I was intending to make the new RNG avoid rounding biases, but we can't assume the interpreter RNG is doing that. We could get a full 32bit number and do it ourselves if necessary (or just use the 32 bit result as the seed if that's safe enough).

proposals/0034-glulx-random.md
curiousdannii: Could you try writing a ChaCha12 implementation in Inform 6 as a POC?

I think a cryptographic level PRNG would very much be overkill. The Xorshift family have been around for 20 years and are very well regarded non-cryptographic PRNGs. The xoshiro128** algorithm I'm proposing here is used in .NET 6. It seems to me to be an ideal choice for Inform 7.

But if your POC works, then it could be made into an extension if someone really needed their Inform story to be resistant to timing attacks and the rest. Personally I can't imagine why that would be needed, but you never know what someone might create in the future.

proposals/0034-glulx-random.md
curiousdannii: It's not known to be a problem in this way. At this stage I'm only looking at changing it for Glulx.

zedlopez:
I had already reported as a bug that I7 lets you *say* you're replacing `random` but nothing happens: [I7-2345](https://inform7.atlassian.net/browse/I7-2435).

Of related interest: [I7-2380](https://inform7.atlassian.net/browse/I7-2380): both the Z-machine and Glulx give bad results if you ask for a random number in a range in excess of half the machine's integer domain. This is inherent in the Z-machine, but warrants documenting in I7 or a runtime error, or both. It needn't be true for Glulx.

also of related interest: [I7-2441](https://inform7.atlassian.net/browse/I7-2441)

...but Dannii has already touched on all these points. at first I didn't see the proposal 'cause it's not in main.

curiousdannii:
@zedlopez The `random` function quite clearly operates on a signed range. Improving `GenerateRandomNumber` could be a good idea, but it's not really in scope for this proposal.

zedlopez:
Not trying to scope-creep; just thought it'd be good to have those issue links here for reference.

proposals/0034-glulx-random.md
hanna-kruppe: It wouldn't be difficult, but would it be sensible? I don't know the last time I might have wanted what `@random n` does for negative n (and if I ever want it, it's easy to get that from the existing `random number between A and B` phrase). I can understand that a VM bytecode format wants to be frugal with opcode numbers and assign some possibly useful behavior to the entire range of inputs (here, one 32-bit integer). But there's no reason to replicate that idiosyncratic design 1:1 into Inform, where phrases are already plentiful and happily issue RTPs for nonsensical inputs. As an author, I might want "number from 0 to N exclusive" (e.g., as a list index) sometimes, and "arbitrary 32 bit number" at other times, but I never want to get one when I expected the other because a variable I expected to be positive turned out to be zero. I really don't see any reason to *not* make these two distinct phrases (and scrap the L1 < 0 case entirely).

proposals/0034-glulx-random.md
hanna-kruppe: To be clear, I don't think any notion of "security" against "attacks" is remotely sensible in an IF context. Anyone worried about that is barking up the wrong tree in the wrong forest, and I won't encourage it by publishing an extension for it. I am exclusively worried about RNG quality, not only today but also for however long the choice of algorithm will live on in published games. The story is more murky than you paint it:

- The original xorshift design is indeed ca. 20 years old, but is not well-regarded today and [already faced serious criticism within two years of first being published](https://www.iro.umontreal.ca/~lecuyer/myftp/papers/xorshift.pdf).
- The variants by Blackman and Vigna which are considered good today are significantly younger, especially the ones with `**` / `++` scramblers that these authors actually recommend today (the earliest date I could find was 2018, stated on [the .NET blog](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/)).
- The quality of these generators, and their peers like PCG or SplitMix64, is not such a settled matter. I won't get into the ridiculous niche internet flame war around this and stand by the assessment that all of these are perfectly fine choices today, but I would not bet on this remaining true for ten or twenty years. Just a few years ago, you could read about how great xorshift128+ is (nobody claims this any more) and how it's used by several JavaScript engines (still true, actually).
- I don't expect there will ever be a surprising discovery of `xoshiro128**` or the like being as pathologically bad in some scenario as the much older RNGs that motivate this proposal. But given the long and ongoing history RNGs being proposed, failing (possibly newly developed) statistical tests, being refined to not fail those tests any more, and/or proponents arguing that failing this or that statistical test won't matter for "real applications", it'd be hubris to think the non-crypto RNGs du jour won't have any flaws left to uncover.

Other languages are in a different position: they ship new versions regularly and generally expect software written in that language to be updated, too. For example, when .NET adopted `xoshiro{128,256}**`, the PR discussion (linked from the .NET blog post mentioned above) focused on (1) it being faster than what they had before and (2) them being able to easily change the algorithm *again* in the future.

In contrast to non-crypto RNGs, block ciphers and friends don't have the luxury of only lasting a few years or hand-waving statistical flaws in the hope that it won't matter. However, that doesn't mean they are *only* useful for cryptographic use cases. On the contrary, when used as engine for a plain old RNG (not bothering with forward secrecy, entropy management, side channel resistance, etc.), you just get a normal RNG with no meaningful security guarantees. but with *more certainty* about its quality and *more robustness* (e.g., block ciphers don't have "bad keys" that one should avoid, while most non-crypto RNGs have "bad seeds" that make them go off the rails). The only real downside is that they're somewhat slower than RNGs optimized to be as fast as possible while still passing the standard statistical tests, but that gap has become smaller with (relatively) newer ciphers like ChaCha and only matters for applications that spend most of their time crunching through terabytes of random numbers (i.e., nothing anyone should use Glulx or Inform for).

proposals/0034-glulx-random.md
```diff
@@ -81,7 +81,12 @@ Considerations:
     call is safe in affected interpreters, or instead whether it should make a
     ranged call using a high prime number, or if any priming is necessary.
 
-3. We need to replace the Inform 6 `random` function rather than introducing a
+3. Unless the new RNG is in predictable mode, it will need to be reseeded from
+    `@random` after restoring or undoing. This means an extra variable will be
+    needed to record whether the RNG is in 'random' or 'predictable' mode, as
+    the RNG's own state would otherwise be indistinguishable.
```
hanna-kruppe, line 87: What does "predictable mode" mean in this context? It's already a somewhat fuzzy notion to me when talking about Glulx terps and their `{random,setrandom}` opcodes, but it's even less clear to me what it's supposed to mean for kit code. I would have expected that this RNG is just reseeded via the terp's `@random` on every restore/restart. Why is that not sufficient? What problem do you want to address with this "RNG mode" business?

proposals/0034-glulx-random.md
hanna-kruppe: Looks like [Git before version 1.3.8](https://github.com/DavidKinder/Git/commit/1f24f3e9b7f71d2ab7344b2405e5b43f8aaf1d4e) always used the C standard library's `rand()` function, whose range and quality is very platform-dependent. It's especially limited on Windows (0..0x8000), but since it returns a non-negative `int`, it can be at most 31 bits on any platform with 32-bit `int` (which includes all common 64-bit platforms). Together with the *other* RNG flaw we know to be common, very low quality in the lowest bits, this makes it very annoying to reliably extract a decent 32-bit number. I'm sure there are decent ways, but I'll have to think about it some more.

proposals/0034-glulx-random.md
curiousdannii: When I've used Xorshift before it has been with a multiplying scrambler.

The Xorshift family don't have any bad values (other than 0), all values form a single cycle.

proposals/0034-glulx-random.md
curiousdannii: Predictable mode refers to seeding the RNG with a specified seed. That allows, for example, everyone to play the same randomly generated dungeon in Kerkerkruip. And while in general we don't want to get the same random results after restoring, some authors might want this to be the case to prevent save scumming.

However, even predictable mode is lost after restoring in Glulx... I think this is one case where a divergence from the Glulx spec would be warranted, because it would be much more useful to have predictable mode surviving restores out of the box, rather than telling users to install something like the Xorshift extension again. Alternatively, we could have three modes: 'true' random, predictable, and preserving predictable. But I think it would be better to just make predictable mode preserving. If you don't want predictable mode to be preserved there's little reason to use it.

proposals/0034-glulx-random.md
curiousdannii: There's also no reason not to implement thr behaviour of the opcode. And just because you can't think of a reason why you wouldn't want the negative version doesn't mean someone else wouldn't. This point in the proposal is just to ensure that no one has reduced options compared to before. If someone was calling `@random` directly rather than the `random` function, they could replace their opcode invocation with a call `random_glulx` (if that's what we called it) to get the same result. I wasn't thinking of exposing the I6 function into I7, it would just be there for anyone using the opcode in I6.

proposals/0034-glulx-random.md
curiousdannii: That's informative, thanks. But unless you think it will be impossible to construct a useful seed (possibly from many calls to `@random`), this level of detailed analysis would probably be more helpful in the forum discussion than in this proposal pull request. The proposal doesn't specify all implementation details, just the broad plan. 🙂

erkyrath:
I see this discussion has already gotten detail-heavy. However, I will present my case anyway:

I think this is a bad tradeoff. I7 should not ship an RNG in game code.

It is true that interpreters can have bad RNG implementations, and some old interpreters will never be updated again. However, interpreters *can* be updated, and the ones in most common use are updated fairly regularly. Game files, on the other hand, are almost never updated. 

Baking RNG code into game files means burdening both the player (larger game files and slower code) and the Inform developers (one more subsystem that must be maintained). The benefit is a questionably small number of players using old interpreters. I don't see that as worthwhile. I7 doesn't have a mandate to patch around every possible interpreter bug, and a bad RNG is just an interpreter bug which takes careful statistical tests to detect.

Another point: the interpreter's native RNG (`@setrandom/@random`) can access high-quality RNG from the OS. This is already configured into Glulxe for MacOS and Windows, as well as Linuxes that support getrandom(). A baked-into-I7 solution can't access that -- except by relying on `@setrandom 0` for the original seed. And then you're exposed to interpreter flaws *anyway*. Plus piling RNGs on top of RNGs even more than is already the case.

To go back to a very old patch: The I6 library has the line

    for (i=1 : i<=100 : i++) j = random(i);

This line was added back in Inform 5, if not earlier. It survives to the present day (see SEED_RANDOM_NUMBER_GENERATOR_R). And for what benefit? Working around an early-1990s Z-machine interpreter which called rand(). How many CPU milliseconds (he said, waving his rhetorical arms grandiosely) have been wasted in this attempt to avoid an interpreter bug which may not have existed in live code for twenty years? I'd say the proposal under discussion is the same idea, and will live as long after any possible value it might provide.

erkyrath:
> However, even predictable mode is lost after restoring in Glulx... I think this is one case where a divergence from the Glulx spec would be warranted, because it would be much more useful to have predictable mode surviving restores out of the box

For what it's worth (and sorry if I seem to be piling on): I don't think this is very useful either.

> Predictable mode refers to seeding the RNG with a specified seed. That allows, for example, everyone to play the same randomly generated dungeon in Kerkerkruip.

That's a situation where the _game_ might consider shipping a baked-in RNG. But it is pretty clearly a rare case; very few games are like Kerkerkruip.

curiousdannii:
@erkyrath  Thanks for your feedback.

> a bad RNG is just an interpreter bug which takes careful statistical tests to detect.

It's much worse than this. If someone calls `random(4)` or `random(8)` we've basically got an [xkcd 221](https://xkcd.com/221/) situation right now. If it wasn't that bad then we could probably leave it, but I think it's far beyond the point of reasonableness.

> Baking RNG code into game files means burdening both the player (larger game files and slower code)

It would be larger, but only very slightly. Almost certainly less than 1kb, probably only a couple hundred bytes.
It would be a little slower. I'd estimate about 15 instructions to get the next 32-bit number (and then more to return a ranged result).

The only alternative to implementing a full RNG that I've thought of, would be that if the upper or middle bytes of `@random 0` are reliable enough we could take two of them and merge them together. That would take 7 instructions to get a 32-bit number. So implementing a full RNG would only be about twice as slow.

Or do you have any other ideas? Because "do nothing" isn't good enough, when the flawed RNG is as bad as it is.

(We can also consider cleaning up that Zcode priming.)

hanna-kruppe:
FWIW, I wrote my detailed comments from a "if this is done, what's the best way to do it?" perspective because I did not, initially, have a strong opinion on whether it should be done. But now I am inclined to agree that "this is a bad tradeoff", partly due to how tricky the details turned out but also due to @erkyrath's points. He mentioned several downsides of the proposal, and I am increasingly skeptical about the upsides, too:

> It's much worse than this. If someone calls `random(4)` or `random(8)` we've basically got an [xkcd 221](https://xkcd.com/221/) situation right now. If it wasn't that bad then we could probably leave it, but I think it's far beyond the point of reasonableness.

These pathological outcomes are objectively pretty bad, but at the same time, they have been around for many years before being noticed. We can speculate as to why (perhaps very few games use the RNG in these especially problematic ways? perhaps IF testing is generally not attuned to detect such problems? etc. etc.), but in any case it does not appear to be have caused game-breaking problems for many people in the past and there's no reason to expect that to change suddenly. These are clearly bugs that needed fixing, and they clearly affect at least one game, but the overall impact is likely much smaller than it appears now that it's in the spotlight.

This isn't a reason to ignore the games that *are* affected, nor the users stuck with buggy terps for various reasons. But updating Inform is not a quick fix for them: it can only help in situations where (1) an old and buggy interpreter is used, **and** (2) the game has been built with future Inform releases that implement the proposal. It will take a while for the latter to even be possible, let alone commonplace. Plus, the intersection of (1)&(2) will always be smaller than (1) alone, and possibly *much* smaller depending on various external factors. As an extreme example, if IE-0007 (double precision floats) is implemented in the same I7 release, and depending on how exactly it is implemented, most games built with that Inform version may not even run on the old terps this proposal aims to address. Both Git and Glulxe only had a single release (last year) between implementing double precision opcodes and fixing their RNG issues.

zedlopez:
I'm persuaded by Andrew's points. The prospect of changing the Inform compiler or the built-in kits to have a better RNG can only improve behavior when a player has a combination of a recently built game and an old terp. And the earliest even *that* benefit could apply is some months after the forthcoming release of Inform, when games built using it begin to appear. I'm assuming the latest release versions of Spatterlight and Gargoyle have updated glulx VMs by the next Inform release (and I'm taking for granted that the IDEs and web interpreter templates will).

So, circa late 2024, there could be a benefit for players who are at least one release behind on Spatterlight or Gargoyle or Lectrote, but are using it to play a brand new game. As time goes on, the number of releases behind they would have to be to see a benefit will continue to grow, and, so the number of potential beneficiaries would continue to shrink. As noted above, it's possible there could be regrets with choice of algorithm. But with it in core Inform, we buy  this continuously shrinking window of potential benefit at the cost of being stuck with the consequences of the re-implementation in the long term.

I still think it's relevant to be able to replace the I6 `random` to facilitate writing an extension like Xorshift without needing a kit, so that an individual author can choose to use it. I think the next release notes should mention why an author might like to do so. But I don't think it makes sense to make a fundamental change to what owns responsibility for random number generation.

(No new points made here... just re-stating things above.)

ganelson:
We seem to be moving to a consensus in which the only action needed is from me? I.e., the ability for users to supply their own drop-in replacement for the I6 "random" function. Is that where we've got to?

curiousdannii:
@ganelson  It would also be good if I7 would reject the multiple argument form of `random(1,2,3,4)` consistently.

There are two platforms where old interpreters are particularly likely to be an issue. First is Linux where if you install Gargoyle through the Debian/Ubuntu where updates can be very delayed. But you would still be able to install it manually yourself.

The other difficult platform is Android. But I don't think we've established for certain if any Android terps do actually have problematic RNGs. As an Android user I guess I'm volunteering to test them myself...

Alternatives to this proposal include:

1. What I suggested above, calling `@random` twice and joining them together, if that gives a reliable enough number and protects against low-bit issues.
2. Inferring the age of an interpreter through an unrelated gestalt like double precision floats, and only doing something in old interpreters.

But when I thought of the second alternative, I realised that with #7 we're already planning for near-future versions of Inform to not run on several-year-old interpreters. (It might be time to revisit whether the interpreter landscape is ready for #7 to be implemented now.) And if that's the case then arguably there's no need to do anything for this problem, and this proposal can just be rejected.

hanna-kruppe:
> What I suggested above, calling `@random` twice and joining them together, if that gives a reliable enough number and protects against low-bit issues.

Not to drift off into details again, but note that the very small range of C `rand()` on Windows (15 bits) means you'd actually need at least three outputs to get a full 32 bits on pre-2022 Git-for-Windows. If you also drop a couple of lower bits because they're very flawed on other terps, you'll probably need four outputs. These are implementation details and only worse by a constant factor, but I wanted to point it out because *once again* the details turn out trickier and uglier and slower (cf. Andrew's concern) than they appear as high-level idea.

drpeterbatesuk:
>@ganelson It would also be good if I7 would reject the multiple argument form of random(1,2,3,4) consistently.

or implement it consistently (it's a neat feature of  canonical I6).

curiousdannii:
> or implement it consistently (it's a neat feature of canonical I6).

Overloading the function like that makes it much harder to re-implement it, and I7 already has other ways of doing it at the I7 level.

curiousdannii:
So with no one else really in favour of this proposal, I'll close it now.

I did however also make a pull request to partially mitigate the problem for older interpreters by priming with a random number of `random()` calls. https://github.com/ganelson/inform/pull/150

PR #40, opened 2024-08-02
IE-0040: Version number kind

curiousdannii:
* Proposal: [IE-0040](https://github.com/ganelson/inform-evolution/blob/main/proposals/0040-version-number-kind.md)
* Authors: Graham Nelson
* Status: Draft
* Related proposals: --
* Implementation: None as yet

## Summary

Adds a `version number` kind, whose constant values can be written `v16`, `v2.4`
or `v2.3.12`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ scale, have since 30 July 2022 been proposed and tracked in this repository.
 The core repository for the language itself is
 [ganelson/inform](https://github.com/ganelson/inform).
 
-## In progress 
+## In progress
 
 Proposal                                                                                                 | Began             | Comments
 -------------------------------------------------------------------------------------------------------- | ----------------- | --------


curiousdannii:
@ganelson Proposal looks good. Only issue I see is with version 0: if the version number is meant to be a semver version, then strictly speaking no version 0s can reliably considered compatible. However in the npm and Rust worlds, 0.y acts like x.y, meaning 0.1.* are compatible, and 0.2.* are compatible, but not with each other.

Considering that historically we haven't used version 0 in Inform 7 extensions, I don't think the compatibility phrase needs this special case, but it's probably worth documenting that there's no special behaviour for version 0.

ganelson:
Well, I think you're referring to the slightly strange rule 4: "Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable." This is somewhat unlike the other rules, in that it's more of a social practice than a technical requirement. And rule 11, on precedence, does not exempt major version 0 from its scheme, so semver does require "0.7.7" to precede "0.7.18", say, rather than to be incomparable with it.

We may not have extensions with version numbers 0.*.*, but Glk is still 0.7.5, and dumb-glulxe has interpreter version 0.6.0, so major version 0 is something which will certainly arise for us. Not all version numbers are semantic in the sense of semver.

In any case, Basic Inform doesn't tell people how to use the version number kind, it only makes the kind available: the compatibility code in the documentation is just an example of what people might or might not choose to do.

PR #42, opened 2026-04-10
IE-0042: Release along with separated resources

curiousdannii:
* Proposal: [IE-0042](https://github.com/ganelson/inform-evolution/blob/main/proposals/0042-release-separated-resources.md)
* Authors: Graham Nelson
* Status: Accepted
* Related proposals: None

## Summary

A new `Release along with...` option which places image and sound files outside of a blorbed story file, rather than inside, as is the default; and writes a corresponding resource map file, in order to correlate them to their resource numbers.

### DIFF
--- a/proposals/0042-release-separated-resources.md
+++ b/proposals/0042-release-separated-resources.md
@@ -1,4 +1,4 @@
-# (IE-0042) Release along with separated resources
+# (IE-0042) Release along with separated resources 
 
 * Proposal: [IE-0042](0042-release-separated-resources.md)
 * Discussion PR link: [#42](https://github.com/ganelson/inform-evolution/pull/42)


curiousdannii:
https://intfiction.org/t/adding-mp3-support-to-the-blorb-specification/80260

@erkyrath has been supportive of adding MP3s to Blorbs officially, though the spec hasn't been updated yet. I think we can switch Inform to output them as a `MP3 ` chunk rather than `BINA`.

erkyrath:
Whoops, yeah, I gotta move that forward.

ganelson:
Thanks, @erkyrath - that's helpful. I have revised Inform to output these new MP3 chunks, and updated the proposal. For the time being, Inblorb issues a warning if people release with these, to the effect that the feature is new. At any rate, it should be good enough to experiment with.

PR #45, opened 2026-05-24
IE-0045: Text substitutions with text literals

curiousdannii:
* Proposal: [IE-0045](https://github.com/ganelson/inform-evolution/edit/main/proposals/0045-text-substitutions-with-text-literals.md)
* Authors: Graham Nelson
* Related proposals: --

## Summary

This change to Inform's lexical syntax enables text literals to appear inside
text substitutions for the first time, making it possible for a substitution to
use a phrase which takes text arguments. Since those text literals can
themselves contain substitutions, substitutions can for the first time be nested.


nightpool:
Is there a substantial reason why nested quotes were not just allowed? wouldn't it simply be changing something that was previously a syntax error into valid syntax. This seems naively like it would head off a lot of the issues with backwards incompatible changes that backticks bring.

ganelson:
After a lot of alternatives were tried out, backticks seemed the least disruptive change, among those which were legible at all.

Nested quotes, `"like this [and "this"]"`, would be trickier to parse, but could be done. However, it would break all existing I7 syntax-colouring, and when typed incorrectly would produce inscrutable problem messages - e.g. that the source text ended inside quoted text. For example, a failure to close square brackets correctly, e.g., by:

```
The marble tomb is here. "Your eyes are drawn to [the tomb."
```

which is currently reported as a problem message because of the mismatched `[`, would likely then be read as a huge text substitution running to the end of the file. That would then throw a problem, but quite a hard one to diagnose.

My belief is that backticks have not generally been used in the wording of text substitutions in the wild; given that, I don't think there will be any major issue with backwards incompatible changes.

